PR #81, opened 2014-07-11
Create indices on GIS types as a constraint on the parent object

lmb:
Right now, using Geoalchemy2 together with alembic is problematic, since ga2 creates indices by directly emitting SQL during the DDL generation stage. alembic does not pick this up obviously, and constantly tries to delete the spatial indices when generating migrations.

This pull request moves index creation into the GIS types themselves, making sure that the proper indices appear in the table / model metadata, which is what alembic uses to create migration scripts.

Let me know if I haven't followed your code guidelines fully. I ran all tests and they checked out.

### DIFF
--- a/TEST.rst
+++ b/TEST.rst
@@ -28,10 +28,10 @@ Create the ``gis`` database::
 
     $ sudo -u postgres createdb -E UTF-8 gis
     $ sudo -u postgres psql -d gis -c 'CREATE SCHEMA gis;'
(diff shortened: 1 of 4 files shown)


elemoine:
Thanks for creating this patch. Given your last comment I am going to close this. The issue is tracked here: https://github.com/geoalchemy/geoalchemy2/issues/137.

PR #92, opened 2015-02-17
Read geometries with ST_AsEWKB instead of ST_AsBinary

fredj:
Otherwise geos can't correctly read 3d geometries: geos supports EWKB, not WKB [1]

[1] http://lists.osgeo.org/pipermail/geos-devel/2013-December/006755.html

### DIFF
--- a/geoalchemy2/types.py
+++ b/geoalchemy2/types.py
@@ -23,7 +23,7 @@ class _GISType(UserDefinedType):
 
     This class defines ``bind_expression`` and ``column_expression`` methods
     that wrap column expressions in ``ST_GeomFromEWKT``, ``ST_GeogFromText``,
-    or ``ST_AsBinary`` calls.
+    or ``ST_AsEWKB`` calls.
 
     This class also defines ``result_processor`` and ``bind_processor``
     methods. The function returned by ``result_processor`` converts WKB values
@@ -100,7 +100,7 @@ class _GISType(UserDefinedType):
         return '%s(%s,%d)' % (self.name, self.geometry_type, self.srid)
 
     def column_expression(self, col):
-        return func.ST_AsBinary(col, type_=self)
+        return func.ST_AsEWKB(col, type_=self)
 
     def result_processor(self, dialect, coltype):
         def process(value):
@@ -160,6 +160,11 @@ class Geography(_GISType):
     from_text = 'ST_GeogFromText'
     """ The ``FromText`` geography constructor. Used by the parent class'
         ``bind_expression`` method. """
+
+    def column_expression(self, col):
+        # load column as WKB because ST_AsEWKB is not defined for geography
+        # see http://postgis.net/docs/ST_AsEWKB.html
+        return func.ST_AsBinary(col, type_=self)
 
 
 class Raster(UserDefinedType):
--- a/tests/test_comparator.py
+++ b/tests/test_comparator.py
@@ -95,7 +95,7 @@ class TestOperator():
         eq_sql(expr, '"table".geom <-> ST_GeomFromEWKT(:geom_1)')
         s = geometry_table.select().order_by(
             geometry_table.c.geom.distance_centroid('POINT(1 2)')).limit(10)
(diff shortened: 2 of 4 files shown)


loicgasser:
+1

fredj:
`ST_AsEWKB` is not defined for geography types: http://postgis.net/docs/ST_AsEWKB.html

tsauerwein:
Maybe add an option `ewkb` to `geoalchemy2.types.Geometry` and only use EWKB if that flag is set? Not sure if this will work.

fredj:
`Geography` type fixed by redefining the `column_expression` function

fredj:
cc @elemoine

elemoine:
Sorry, I have some catch-up work on GeoAlchemy. I'll try to review PRs this week.

fredj:
Thanks

elemoine:
If we read geometries using `ST_AsEWKB` instead of `ST_AsBinary` then it means that the `WKBElement` objects created by the [`result_processor`](https://github.com/geoalchemy/geoalchemy2/blob/master/geoalchemy2/types.py#L105-L109) will wrap an EWKB value instead of an WKB value. But `WKBElement` [uses](https://github.com/geoalchemy/geoalchemy2/blob/master/geoalchemy2/elements.py#L98-L103) `ST_GeomFromWKB` when used in an SQL expression, which will result in a warning being generated by PostGIS. For example:

```
gis=# SELECT ST_GeomFromWKB(ST_AsEWKB(ST_GeomFromEWKT('SRID=4326;POINT(1 2)')));
WARNING:  OGC WKB expected, EWKB provided - use GeometryFromEWKB() for this
                   st_geomfromwkb                   
----------------------------------------------------
 0101000020E6100000000000000000F03F0000000000000040
(1 row)
```

One option may be to create an `EWKBElement` class and have the `result_processor` create `EWKBElement` objects rather than `WKBElement` objects.

geoalchemy2/types.py
```diff
@@ -161,6 +161,11 @@ class Geography(_GISType):
     """ The ``FromText`` geography constructor. Used by the parent class'
         ``bind_expression`` method. """
 
+    def column_expression(self, col):
+        # load column as WKB because ST_AsEWKB is not defined for geography
+        # see http://postgis.net/docs/ST_AsEWKB.html
+        return func.ST_AsBinary(col, type_=self)
```
elemoine: Instead of overriding the `column_expression` function it would be more consistent to introduce an `as_binary` property, as done with [`from_text`](https://github.com/geoalchemy/geoalchemy2/blob/master/geoalchemy2/types.py#L83-L85).

elemoine:
> Maybe add an option ewkb to geoalchemy2.types.Geometry and only use EWKB if that flag is set? Not sure if this will work.

I like the idea of adding an `use_ewkb` flag given that using `ST_AsEWKB` is not compatilble with the `WKBElement` class (which uses `ST_GeomFromWKB`).

elemoine:
Closing this due to lack of activity. @fredj please reopen if you want to discuss this again.

loicgasser:
As I'd like to propose a PR for that, if I understand correctly, we could:

Add a flag `use_ekwb` to `_GISType` which will set a string let's say `self.as_binary` during initialization and which can be then called via `column_expression`  using:
`return getattr(func, self.as_binary)(col, type_=self)`

same in result processor, but should I create a EWKBElement or extend WKBElement using the same flag (`use_ewkb`)

@elemoine @tsauerwein What do you think? Does it sound like a plan?

lukas-gitl:
Is this going to brake the hack-fix for issue #93?

elemoine:
> Is this going to brake the hack-fix for issue #93?

No, as nothing will change for you if don't set `use_ewkb` to `True`. I looked at #93 when you created it, but I did not have time to investigate the problem. We can continue the discussion in #93 and attempt to find a proper solution if you have time.

lukas-gitl:
That sounds great! Thank you @elemoine

PR #95, opened 2015-05-04
Adds basic equals comparator to basic spatial type

lukecampbell:
- This addresses an error raised when SQLAlchemy 1.0+ attempts to invoke
  a comparator on the WKB Elements with a "None", which one might desire
  in order to update a column to NULL.

### DIFF
--- a/geoalchemy2/elements.py
+++ b/geoalchemy2/elements.py
@@ -53,6 +53,14 @@ class _SpatialElement(object):
         func_ = functions._FunctionGenerator(expr=self)
         return getattr(func_, name)
 
+    def __eq__(self, other):
+        '''
+        Basic equals comparator so that values can be set to NULL
+        '''
+        if isinstance(other, _SpatialElement):
+            return self.srid == other.srid and self.data == other.data
+        return False
+
 
 class WKTElement(_SpatialElement, functions.Function):
     """
--- a/tests/test_elements.py
+++ b/tests/test_elements.py
@@ -37,6 +37,27 @@ class TestWKTElement():
             u'ST_GeomFromText_2': -1
         }
 
+    def test_compare(self):
+        # Two identical points
+        e0 = WKTElement('POINT(1 2)')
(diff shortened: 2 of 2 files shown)


Turbo87:
is there some sort of unit test that could verify the desired behavior?

lukecampbell:
I had an issue with the latest changes in SQLAlchemy integrating due to a comparison.
http://docs.sqlalchemy.org/en/latest/changelog/migration_06.html#expression-language-changes

There's an internal comparison done when an update occurs and it attempts to compare a WKB to the new update value. The issue this addresses is the situation where the update value is not a Spatial element type. This will ultimately raise a TypeError, and the default comparator doesn't really work when they are both spatial operators. 

I'm not intimately familiar with how comparison _should_ be done but I took an initial stab at it.

lukecampbell:
I could write a unit test at some point. I'm ok with holding off a merge until it's tested. Probably better that way.

Turbo87:
hmm, in theory your link shouldn't be relevant since this project is targeting SQLAlchemy v0.8 and above and your link leads to a migration guide for v0.6...

Turbo87:
that would be awesome, thanks

lukecampbell:
I agree, in theory it shouldn't matter but at some point this happened:
https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/orm/persistence.py#L457-L459

Which was the source of a lot of pain for me. In older versions of SQLAlchemy, I'm assuming, this expression evaluated to a simple true or false. Without a proper comparator defined, it raises a TypeError

https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/sql/elements.py#L2722-L2726

I just went through the painstaking task of upgrading from SQLAlchemy 1.0 to the latest to bring in changes that will help us support full text search and some other stuff.

elemoine:
This probably did not get the attention it deserved. I am sorry about that. Is this issue still relevant. An example demonstrating the problem would be great.

PR #100, opened 2015-05-07
Travis cleanup

tigerfoot:
This is a proposal to cleanup a bit the travis build.
Add python 3x support
Remove old SQLAlchemy 0.8x
Add SQLAlchemy 1.0.3


Turbo87:
why did you add the `psycopg2` dependency?

Turbo87:
Unfortunately I am still running SQLAlchemy v0.8 on one of my projects so I would like to keep those tests for now. I would be okay with dropping Python 2.6 but @elemoine would have to agree on that. Python 3.3 was already tested, so "adding" it sounds strange?!

tigerfoot:
@Turbo87 the fact is if we want to make Geoalchemy2 python3 ready, we have to use underlying lib that works too. I've a big doubt about 0.8 
Python 2.6 is upstream dead but it's still up to you.
So to move forwards, can we imagine keeping travis build for 0.3 but only for python3 ?
Or what would be you idea ?

Jenselme:
From what I understand from [the documentation](http://docs.sqlalchemy.org/en/rel_0_9/changelog/migration_09.html), 0.8 is only partially compatible with python3 and the full compatibility comes with 0.9

Turbo87:
how about running something like the following matrix?

- Py 2.7 with SA 0.8, 0.9, 1.0
- Py 3.3 with SA 0.9, 1.0
- Py 3.4 with SA 0.9, 1.0

do we really need to drop SA 0.8 support on Py2.x just to make this lib compatible with Py3.x?

tigerfoot:
This plan make sense to me.
So people that still want to use SA 0.8 can, and we got a test suite that doesn't block the project to move forward.

tigerfoot:
@Turbo87 could you have a look here 
https://github.com/tigerfoot/geoalchemy2/tree/travis-cleanup2
I've redone the job with your remarks. If okay I will superseed this by a new pr

Turbo87:
@tigerfoot you can reuse this PR if you want to: `git push REMOTE travis-cleanup2:travis-cleanup --force`

tigerfoot:
done.

Turbo87:
cool, thanks!

elemoine:
Thanks for these changes. They make sense to me.

PR #102, opened 2015-05-24
Rename tox travis env SQLAlchemy 103 to 10

tigerfoot:
* Modifiy SA 103 to 10 add support for all SA 1.0 version
  * Adjust tox to last SA 1.0.4 version
Should be more appropriate on moving target like SA


elemoine:
LGTM, even though SQLA 1.0.5 has been released :)

Issue #101, opened 2015-05-09
Use TravisCI to publish PyPI packages


Turbo87:
I'm doing this on my own projects too, and it works pretty well.

see http://www.agmweb.ca/2015-05-08-travis-pypi

@elemoine since the package on PyPI is registered to your account you would have to install this.

Turbo87:
@elemoine any comments?

elemoine:
Sounds very good. I'll look at it. Again, sorry for the delay responding.

PR #104, opened 2015-06-09
Use TravisCI to publish PyPI packages

elemoine:
Note that this is untested. I need to merge it to be able to test it.

Fixes https://github.com/geoalchemy/geoalchemy2/issues/101.


elemoine:
It seems to work. GeoAlchemy2-0.2.5 just landed in pypi: https://pypi.python.org/simple/geoalchemy2/.

Turbo87:
awesome :+1:

PR #114, opened 2015-10-18
Distribute GeoAlchemy as wheels

elemoine:
This PR makes the Travis build create a wheel and upload it to pypi when a tag/release is created.


elemoine:
Ok merging this. If people agree we can release 0.2.6, with a wheel. For The Wheels :)

sbrunner:
Nice :-)

PR #121, opened 2015-11-03
Add EWKB support

loicgasser:
This is an attempt to integrate what has been discussed in https://github.com/geoalchemy/geoalchemy2/pull/92

The reason: http://osgeo-org.1560.x6.nabble.com/WKB-representation-is-EWKB-not-OGC-td5092794.html
(e.g. Shapely relies on EWKB, which is fine in 2D, every WKB == every EWKB, but when it comes to 3D geoms it is not the case, so we can't decode 3D geoms coming from postgis with Shapely using the current geoalchemy setup...)

It adds one new option to `Geometry`, namely `use_ewkb` and one new element `EWKBElement`.

I had to change the `GenericFunction` to make sure the instance of the `type` was used to take into account the `use_ewkb` option.
Relevant code in SQLAlchemy can be found here:
https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/sql/functions.py#L529

For EWKB postgis formats:
http://postgis.net/docs/ST_AsEWKB.html
http://postgis.net/docs/manual-2.2/ST_GeomFromEWKB.html

Now, I am not 100% sure that's a the design you want, so let me know and I'll adapt gladly.

### DIFF
--- a/geoalchemy2/elements.py
+++ b/geoalchemy2/elements.py
@@ -113,6 +113,35 @@ class WKBElement(_SpatialElement, functions.Function):
         return binascii.hexlify(self.data)
 
 
+class EWKBElement(_SpatialElement, functions.Function):
+    """
+    Instances of this class wrap a EWKB value. Geometry values read
+    from the database are converted to instances of this type. You may
+    need to create instances of ``EWKBElement`` yourself.
+
+    Note: you cannot create ``EWKBElement`` objects directly from Shapely
+    geometries using the :func:`geoalchemy2.shape.from_shape` function as there
+    is currently no support in Shapely for SRIDs. Instead, you must have a
+    valid EWKB (e.g. beggining with SRID=4326;) and use directly
+    a ``EWKBElement`` (e.g. ``EWKBElement(buffer(ewkb))``).
+    """
+
+    def __init__(self, *args, **kwargs):
+        _SpatialElement.__init__(self, *args, **kwargs)
+        functions.Function.__init__(
+            self,
+            "ST_GeomFromEWKB",
+            self.data
+        )
+
+    @property
+    def desc(self):
+        """
+        This element's description string.
+        """
+        return binascii.hexlify(self.data)
+
+
 class RasterElement(FunctionElement):
     """
     Instances of this class wrap a ``raster`` value. Raster values read
--- a/geoalchemy2/shape.py
+++ b/geoalchemy2/shape.py
@@ -7,10 +7,12 @@ This module provides utility functions for integrating with Shapely.
     functions of this module have to ensure that `Shapely` is available.
 """
 
+from copy import copy
 import shapely.wkb
 import shapely.wkt
+import shapely.geos
 
-from .elements import WKBElement, WKTElement
+from .elements import WKBElement, EWKBElement, WKTElement
 from .compat import buffer, bytes
 
 
@@ -24,17 +26,18 @@ def to_shape(element):
         lake = Session.query(Lake).get(1)
         polygon = to_shape(lake.geom)
     """
-    assert isinstance(element, (WKBElement, WKTElement))
-    if isinstance(element, WKBElement):
+    assert isinstance(element, (WKBElement, EWKBElement, WKTElement))
+    if isinstance(element, WKBElement) or isinstance(element, EWKBElement):
         return shapely.wkb.loads(bytes(element.data))
     elif isinstance(element, WKTElement):
         return shapely.wkt.loads(element.data)
 
 
-def from_shape(shape, srid=-1):
+def from_shape(shape, srid=-1, use_ewkb=False):
     """
     Function to convert a Shapely geometry to a
-    :class:`geoalchemy2.types.WKBElement`.
+    :class:`geoalchemy2.types.WKBElement` or a
+    :class:`geoalchemy2.types.EWKBElement.
 
     Additional arguments:
 
@@ -43,9 +46,27 @@ def from_shape(shape, srid=-1):
         An integer representing the spatial reference system. E.g. 4326.
         Default value is -1, which means no/unknown reference system.
 
+    ``use_ewkb``
+
+        Boolean indicating whether converting the shape to
+        a WKBElement or to a EWKBElement. srid is ignored if use_ewkb is set.
+        Default value to False.
+
     Example::
 
         from shapely.geometry import Point
         wkb_element = from_shape(Point(5, 45), srid=4326)
     """
-    return WKBElement(buffer(shape.wkb), srid=srid)
+    if use_ewkb:
+        wkbwriter_defaults = shapely.geos.WKBWriter.defaults
+        wkbwriter_defaults_copy = copy(wkbwriter_defaults)
+        wkbwriter_defaults_copy['include_srid'] = True
+        shapely.geos.WKBWriter.defaults = wkbwriter_defaults_copy
+        try:
+            element = EWKBElement(buffer(shape.wkb))
+        finally:
+            shapely.geos.WKBWriter.defaults = wkbwriter_defaults
+        return element
+    else:
+        element = WKBElement(buffer(shape.wkb), srid=srid)
+    return element
--- a/geoalchemy2/types.py
+++ b/geoalchemy2/types.py
@@ -13,7 +13,9 @@ from sqlalchemy.dialects import postgresql
 from sqlalchemy.dialects.postgresql.base import ischema_names
 
 from .comparator import BaseComparator, Comparator
-from .elements import WKBElement, WKTElement, RasterElement, CompositeElement
+from .elements import (
+    WKBElement, EWKBElement, WKTElement, RasterElement, CompositeElement
+)
 
 
 class _GISType(UserDefinedType):
@@ -84,6 +86,10 @@ class _GISType(UserDefinedType):
     """ The name of ST_*FromText function for this type.
         Set in subclasses. """
 
+    as_binary = None
+    """ The name of ST_*AsBinary function for this type.
+        Set in subclasses. """
+
     comparator_factory = Comparator
     """ This is the way by which spatial operators are defined for
         geometry/geography columns. """
@@ -100,7 +106,7 @@ class _GISType(UserDefinedType):
         return '%s(%s,%d)' % (self.name, self.geometry_type, self.srid)
 
     def column_expression(self, col):
-        return func.ST_AsBinary(col, type_=self)
+        return getattr(func, self.as_binary)(col, type_=self)
 
     def result_processor(self, dialect, coltype):
         def process(value):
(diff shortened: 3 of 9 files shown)


geoalchemy2/elements.py
```diff
@@ -113,6 +113,35 @@ def desc(self):
         return binascii.hexlify(self.data)
 
 
+class EWKBElement(_SpatialElement, functions.Function):
+    """
+    Instances of this class wrap a EWKB value. Geometry values read
+    from the database are converted to instances of this type. You may
+    need to create instances of ``EWKBElement`` yourself.
```
elemoine: The `WKBElement` docs say that "In most cases you won't need to create WKBElement instances". Why do the `EWKBElement` docs say something different?

geoalchemy2/elements.py
```diff
@@ -113,6 +113,35 @@ def desc(self):
...
+
+    Note: you cannot create ``EWKBElement`` objects directly from Shapely
```
elemoine: I would not mention Shapely here. The `shape` module is the only place where Shapely should be mentioned in my opinion. Remember that using Shapely with GeoAlchemy is totally optional.

geoalchemy2/elements.py
```diff
@@ -113,6 +113,35 @@ def desc(self):
...
+    geometries using the :func:`geoalchemy2.shape.from_shape` function as there
+    is currently no support in Shapely for SRIDs. Instead, you must have a
+    valid EWKB (e.g. beggining with SRID=4326;) and use directly
+    a ``EWKBElement`` (e.g. ``EWKB(buffer(ewkb))``).
```
elemoine: What's `EWKB` in the example?

geoalchemy2/elements.py
elemoine: I now realize that the `WKBElement` does mention Shapely. So I guess you can ignore the above the comment.

geoalchemy2/elements.py
elemoine: Also, isn't this entire note in opposition to your changes to `from_shape`? It looks to me that with your changes the `from_shape` function does support creating `EWKBElement`. No?

geoalchemy2/shape.py
```diff
@@ -24,14 +24,19 @@ def to_shape(element):
...
         polygon = to_shape(lake.geom)
     """
-    assert isinstance(element, (WKBElement, WKTElement))
+    assert isinstance(element, (WKBElement, EWKBElement, WKTElement))
     if isinstance(element, WKBElement):
         return shapely.wkb.loads(bytes(element.data))
+    elif isinstance(element, EWKBElement):
+        # This tells shapely to include SRIDs
+        from shapely import geos
+        geos.WKBWriter.defaults['include_srid'] = True
```
elemoine: This looks like a global setting.

geoalchemy2/shape.py
```diff
@@ -24,14 +24,19 @@ def to_shape(element):
...
+        return shapely.wkb.loads(bytes(element.data))
     elif isinstance(element, WKTElement):
         return shapely.wkt.loads(element.data)
 
 
-def from_shape(shape, srid=-1):
+def from_shape(shape, srid=-1, use_ewkb=False):
     """
     Function to convert a Shapely geometry to a
     :class:`geoalchemy2.types.WKBElement`.
```
elemoine: We want to extend this sentence.

geoalchemy2/shape.py
```diff
@@ -43,9 +48,18 @@ def from_shape(shape, srid=-1):
         An integer representing the spatial reference system. E.g. 4326.
         Default value is -1, which means no/unknown reference system.
 
+    ``use_ewkb``
+
+        A boolean to indicate whether to use EWKB or WKB.
```
elemoine: -> Boolean indicating whether converting the shape to a `WKBElement` or to a `EWKBElement`.

geoalchemy2/shape.py
```diff
@@ -43,9 +48,18 @@ def from_shape(shape, srid=-1):
...
+        In EWKB the SRID is included, as a result srid parameter is ignored.
+        Default value is False, which means that WKB is used.
```
elemoine: I think you can replace this with "`srid` is ignored if `use_ewkb` is set."

geoalchemy2/shape.py
elemoine: It's also strange that you configure `WKBWriter` while `to_shape` uses `shapely.wkb.loads`, which is about reading. No?

geoalchemy2/types.py
```diff
@@ -84,28 +86,43 @@ class _GISType(UserDefinedType):
...
         self.spatial_index = spatial_index
         self.management = management
+        self.use_ewkb = use_ewkb
+        if self.as_binary is not None:
+            self.as_binary = self.as_binary
+        else:
+            if self.use_ewkb:
+                self.as_binary = 'ST_AsEWKB'
+            else:
+                self.as_binary = 'ST_AsBinary'
```
elemoine: How about changing the whole block to the following?

```py
        if not self.as_binary:
            self.as_binary = 'ST_AsEWKB' if self.use_ewkb else 'ST_AsBinary'
```

geoalchemy2/types.py
```diff
@@ -84,28 +86,43 @@ class _GISType(UserDefinedType):
...
+    """ The name of ST_*AsBinary function for this type.
+        Set using the option ``use_ewkb`` or in subclasses. """
+
     comparator_factory = Comparator
     """ This is the way by which spatial operators are defined for
         geometry/geography columns. """
 
     def __init__(self, geometry_type='GEOMETRY', srid=-1, dimension=2,
-                 spatial_index=True, management=False):
+                 spatial_index=True, management=False, use_ewkb=False):
```
elemoine: Thinking about it, the `use_ewkb` constructor argument should not be on `_GISType`, it should be on `Geometry` instead.

tests/test_types.py
```diff
@@ -13,13 +13,22 @@ def eq_sql(a, b):
 
 @pytest.fixture
 def geometry_table():
-    table = Table('table', MetaData(), Column('geom', Geometry))
+    table = Table('table', MetaData(), Column('geom', Geometry()))
```
elemoine: Any specific reason for changing this?

tests/test_types.py
```diff
@@ -13,13 +13,22 @@ def eq_sql(a, b):
...
 
+@pytest.fixture
+def geometry_table_ewkb():
+    table = Table('table', MetaData(), Column('geom', Geometry(use_ewkb=True)))
+    return table
 
 @pytest.fixture
 def geography_table():
-    table = Table('table', MetaData(), Column('geom', Geography))
+    table = Table('table', MetaData(), Column('geom', Geography()))
```
elemoine: Again, any reason for changing this line?

geoalchemy2/shape.py
```diff
@@ -43,9 +48,18 @@ def from_shape(shape, srid=-1):
...
+
     Example::
 
         from shapely.geometry import Point
         wkb_element = from_shape(Point(5, 45), srid=4326)
     """
+    if use_ewkb:
+        return EWKBElement(buffer(shape.wkb))
```
elemoine: This is probably where you actually want to change the `WKBWriter` configuration. Something like the following:

```py
    if use_ewkb:
        wkbwriter_defaults = geos.WKBWriter.defaults
        wkbwriter_defaults_copy = copy(wkbwriter_defaults)
        wkbwriter_defaults_copy['include_srid'] = True
        geos.WKBWriter.defaults = wkbwriter_defaults_copy
        try:
            element = EWKBElement(buffer(shape.wkb))
        finally:
            geos.WKBWriter.defaults = wkbwriter_defaults
    else:
        element = WKBElement(buffer(shape.wkb), srid=srid)
    return element
```

But I think what we really want is add an `ewkb` to Shapely and do:

```py
    if use_ewkb:
        return EWKBElement(buffer(shape.ewkb))
    return WKBElement(buffer(shape.wkb), srid=srid)
```

elemoine:
I think it looks good. And I guess my main concern is about the global `WKBWriter` setting. I'd suggest that we create a PR to Shapely to suggest adding an `ewkb` property. It would be much better. What do you think?

elemoine:
Also, I am wondering about totally replacing WKB by EWKB in the future. Why would we want to keep WKB? What can we do with WKB that we can't do with EWKB?

What we could possibly do: release 0.3.0 with `use_ewkb` set to `False` by default (as done in your patch), and recommend that people set it to `True` in their applications. Then set `use_ewkb` to `True` by default in 0.4.0, and finally remove `WKBElement` and assiocated code in 0.5.0 (or 1.0).

What do you think?

loicgasser:
Hi, thank you for taking the time to have a look at it.

Why would we want to keep WKB? What can we do with WKB that we can't do with EWKB?

Despite what GEOS is advertising here: http://geos.osgeo.org/doxygen/classgeos_1_1io_1_1WKBReader.html
they actually use a partial implementation of Postgis EWKB specs which means that we would indeed be better of using only EWKB (no support for 3D using geoalchemy2 at the moment without using a hack). Now my may concern is related to the `from_shape` function in conjuction with the SRID enforcements in DB. As `ST_GeomFromEWKB` only takes one argument, we have to prepend the SRID in the EWKB, which we can't do at the moment with Shapely. Having a function in Shapely like `shape.ewkb(data, srid)` would solve all our problems. `EWKBElement(buffer(shape.ewkb(srid))))` (if srid is None then simply to prepend any). This is for the writing part.

Now for the reading part and the `to_shape` function, I am pretty sure GEOS creates the correct Geometry and that Shapely simply leaves out the SRID part, so it's not a problem I think.

I'll have some time tomorrow to make some corrections and have a closer look at your comments, maybe add a few tests and come up with an update on this patch for a 0.3.0 release (a good thing if we introduce it step by step). I agree that we should propose a patch in Shapely. Might be hard as really they don't care about SRIDs and it would be a first...

elemoine:
> Now for the reading part and the to_shape function, I am pretty sure GEOS creates the correct GEOS Geometry with the SRID and that Shapely simply doesn't expose the SRID part.

Yes, it's also my understanding.

What I don't understand is why your `to_shape` function configures the `WKBWriter` class, as this class is not involved in the reading process.

geoalchemy2/shape.py
elemoine: I now understand that the above code (with `wkbwriter_defaults` and stuff) won't work because the srid is not passed to the WKBWriter.

elemoine:
> Shapely relies on EWKB

This is what we read in the issue description. Did you mean to write "WKB" here? If so please udpate the issue description.

elemoine:
@loicgasser, what do you think about https://github.com/geoalchemy/geoalchemy2/compare/master...elemoine:st_ewkb? This is based on @fredj's initial patch, with additional code to address the issue I mentioned in https://github.com/geoalchemy/geoalchemy2/pull/92#issuecomment-78916246. Is it too simplistic for your case? Am I missing the main point of your patch?

loicgasser:
> What I don't understand is why your to_shape function configures the WKBWriter class, as this class is not involved in the reading process.
That was a mistake, just updated with your solution.

loicgasser:
> This is what we read in the issue description. Did you mean to write "WKB" here? If so please udpate the issue description.
Yes sorry

loicgasser:
> what do you think about master...elemoine:st_ewkb?
Probably makes more sense to use extended because we don't really change the format.

elemoine:
@loicgasser I added a Shapely/EWKB integration test to my branch.

loicgasser:
I think your solution looks better mitigates all the issues and may simply be easier to use overall.

loicgasser:
I'll close and let you open a new PR with your work. Thank you very much

PR #125, opened 2015-11-06
Fix flake8 in tests

loicgasser:
Just noticed that, I think you may want to run the tests through flake8 as well


elemoine:
Thanks.

Issue #50, opened 2013-05-13
Can't create new table with Geometry, "Error type modeified for geometry not allowed"


tomasholderness:
I have to say I'm impressed with GeoAlchemy2. Unfortunately, when I try and create a table with a geometry column I get the following error:

"sqlalchemy.exc.ProgrammingError: (ProgrammingError) type modifier is not allowed for type "geometry"
LINE 4:  the_geom geometry(GEOMETRY,-1)"

Any ideas how I can fix this?

I'm using PostGIS 1.5 (which I guess might be the problem?),PostgreSQL 9.1 and GeoAlchemy2 0.2 

Thanks.

PR #126, opened 2016-02-02
Rewrite unit test instruction to be easier to use.

adroffner:
Combine all the "gis" test database SQL into a single PSQL script.

Modify PostgreSQL test connection to report the port number.
Some users may need alternate ports, e.g. 5433, to run postgres.

### DIFF
--- a/TEST.rst
+++ b/TEST.rst
@@ -1,6 +1,15 @@
-=====
-Tests
-=====
+=============================
+Tests using the pytest runner
+=============================
+
+QUICK TEST
+==========
+
+Are your Python, Postgres and PostGIS dependencies are all installed?
+Run the tests in a created ``gis`` database::
+
+    $ sudo -u postgres psql ./tests/create_database.sql
+    $ py.test
 
 Install system dependencies
 ===========================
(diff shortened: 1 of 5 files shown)


adroffner:
Before I can add my changes to the project, I need to run the unit tests. So, I made changes to the unit test setup. I think this makes it much easier to create the test database named "gis".

Of course, the tests run fine after using my PSQL db-create script.

adroffner:
Apparently, the breaks the PostGIS 1.5 tests. The PostGIS 2.x tests succeed, which is the version I developed my code on.

elemoine:
Thanks for the patch. We need something that works with both PostGIS 1 and 2.

adroffner:
I only have PostGIS 2.1 installed on PG 9.4 server. Can you run PostGIS 1.5?

You'll see the version 1.5 PSQL scripts commented out in
./tests/test_database.sql above CREATE EXTENSION postgis;

The hard part is finding an IF-statement in PSQL syntax to choose 1.5 or
version 2+.

adroffner:
The PostGIS version 1.5 are failing due to a missing debian package for it. Travis CI is raising the error in ".travis.yml" (I'm not skilled with Travis CI). I can't fix this.

MISSING PACKAGES: The versions don't match.
1) postgresql-9.1-postgis=1.5.3-2
2) postgis=1.5.3-2

See this near line #50:

install:
  # Install PostGIS 1.5
  - if [[ "$POSTGIS_VERSION" == "1.5" ]]; then sudo apt-get install -y --force-yes postgresql-9.1-postgis=1.5.3-2 postgis=1.5.3-2; fi

elemoine:
@adroffner is this patch still relevant? The build still fails on Travis.

elemoine:
I am closing this due to lack of activity.

PR #127, opened 2016-02-02
Add a GeoJSON type to handle REST APIs stored in PostGIS

adroffner:
PostGIS supports http://geojson.org/ strings as way to read GEOMETRY columns, like WKT. GeoJSON  features are a good way to pass GIS data via a REST API.

This new GeoAlchemy2 type, GeometryJSON(), lets us marshall GEOMETRY columns to a REST API (nearly) automatically. The GeoJSONElement() wrapper can show it as either a GeoJSON string or a dict object.

### DIFF
--- a/TEST.rst
+++ b/TEST.rst
@@ -1,6 +1,15 @@
-=====
-Tests
-=====
+=============================
+Tests using the pytest runner
+=============================
+
+QUICK TEST
+==========
+
+Are your Python, Postgres and PostGIS dependencies are all installed?
+Run the tests in a created ``gis`` database::
+
+    $ sudo -u postgres psql ./tests/create_database.sql
+    $ py.test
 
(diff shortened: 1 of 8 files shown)


elemoine:
Converting a GeoAlchemy `_SpatialElement` to a GeoJSON object is pretty easy already, using GeoAlchemy's `to_shape` function and the [geojson](https://pypi.python.org/pypi/geojson/1.3.3) library. For example:

```py
import geojson
from geoalchemy2 import WKTElement
from geoalchemy2.shape import to_shape

element = WKTElement('POINT(1 2)')
geojson_str = geojson.dumps(to_shape(element))
```

So I don't think I would ike to add specific objects to GeoAlchemy for that purpose.

I am going to close this for now, but feel free to reopen if you think this deserves more discussion.

PR #129, opened 2016-02-26
WKTElement: fix __str__ under python 3

b11z:
In py3, WKTElement.desc() returns string and WKBElement.desc() returns
bytes, so the __str__ implementation in the subclass needs to handle both.

>  str(WKTElement('POINT(1 2)')
AttributeError("'str' object has no attribute 'decode'")

### DIFF
--- a/geoalchemy2/elements.py
+++ b/geoalchemy2/elements.py
@@ -41,6 +41,8 @@ class _SpatialElement(object):
     def __str__(self):
         if not PY3:
             return self.desc
+        if isinstance(self.desc, str):
+            return self.desc
         return self.desc.decode('utf-8')
 
     def __repr__(self):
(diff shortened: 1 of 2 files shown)


elemoine:
I agree that there is a problem on py3. But I am not sure I agree with the fix. I wonder why `__str__` doesn't just return `self.desc`, both on py2 and py3:

```python
    def __str__(self):
        return self.desc
```

What's the problem with this?

Contrary to what this issue's description says both `WKTElement.desc` and `WKBElement.desc` return a string. So I don't understand why the above `__str__` implementation would not be suitable.

b11z:
I was confused by this too. WKBElement.desc is actually returning bytes, because in python3 binascii.hexlify returns bytes:
```
In [1]: import binascii
In [2]: binascii.hexlify(b'3.14159')
Out[3]: b'332e3134313539'
```
Also check out the TestWKBElement, it asserts that the type is bytes. If modifying the .desc methods to return strings is the better change that sounds good to me too!

elemoine:
Ok thanks @b11z. I understand.

elemoine:
@b11z see https://github.com/geoalchemy/geoalchemy2/compare/master...elemoine:element-str. I'll create a new PR if the builds pass on Travis.

b11z:
@elemoine thanks, that looks good to me!

PR #130, opened 2016-03-11
Fix SpatialElement.__str__ on Python 3

elemoine:
This is an alternative to https://github.com/geoalchemy/geoalchemy2/pull/129, as discussed with @b11z. Thank you @b11z for the initial patch.


sbrunner:
Looks good, thanks @elemoine :-)

PR #138, opened 2016-10-22
Simplify and fix tests

elemoine:
The current tests do not pass with SQLA 1.1.2 (the most recent stable version). This PR makes the tests compatible with SQLA 1.1.2, and removes the execution of the tests on SQLA 0.8 and 0.9 in the CI. The execution of the tests on Python 2.6 is also removed.


elemoine:
I'll merge this soon if there's no push-back.

sbrunner:
Looks good, there is just some coverage error :-)

elemoine:
> Looks good, there is just some coverage error :-)

This is minor. It's related to SQLA < 0.9 not being tested anymore. See https://coveralls.io/builds/8461743/source?filename=geoalchemy2%2Felements.py for example.

sbrunner:
Yes :-)
I don't like a "normal" test failed, than I think that you should add some `pragma: no cover` or change the coveralls config to make the test pass :-)

sbrunner:
Thanks @elemoine  :-)

Issue #137, opened 2016-10-20
Consider defining explicit Index() object instead of creating by hand


richardsmd:
geoalchemy creates spatial indexes by default, which I _love_

This causes an unfortunate interplay with [alembic](http://alembic.zzzcomputing.com/en/latest/) when using the `--autogenerate` command. Basically, alembic is expecting an `Index` instance to be defined somewhere to record the presence of an index.

This can be an explicit declaration or[ the one created by passing `index=True` to a Column](https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/sql/schema.py#L1340). In both these cases, the underlying `metadata` gets updated that the Index exists.

geoalchemy uses sqlalchemy's `event` system to create the index in pure SQL. As a result there is no updating of `metadata` and alembic's `--autogenerate` will behave inappropriately.

The workaround is pretty simple once you figure it out
1. Update calls to geometry e.g., `point = Geometry('POINT', spatial_index=False)`
1. Explicitly create the index - `Index('idx_tablename_point', OrmObject.__table__.c.point, postgresql_using='gist')`

Arguably this is a problem of alembic, or of the user of alembic (who is urged to manually review and edit all `--autogenerate` output before executing). Alternatively, you might feel that the index creation should be updated to use explicit `Index` constructs
- If you think the problem is alembic, what are your thoughts on adding a _for Alembic users_ section to the documentation?
- If you think changing geoalchemy2 to create an explicit `Index` is appropriate, what are you thoughts on accomplishing that? I believe listening for the `'before-create'` event and creating the Index as described above would be sufficient.

Thank you for your time,

Mike

richardsmd:
Example code demonstrating the issue. Note that alembic isn't present at all - this just demonstrates the error that arises when both sqlalchemy and geoalchemy try and create the same index if one defines an Index manually without explicitly setting `spatial_index=False`

```python
#!/usr/bin/env python

import psycopg2

from sqlalchemy import (
    create_engine,
    MetaData,

    Column,
    Index,

    Integer,
)

from sqlalchemy.ext.declarative import declarative_base

from geoalchemy2 import Geometry


def connect():
    """
    Connects via 'peer' authentication to the local postgres instance for my account.

    Modify this to connect as you want. DB must already have PostGIS extension.
    """
    user = 'richardsmd'
    return psycopg2.connect(user=user, database=user)

engine = create_engine('postgresql://', creator=connect)

meta = MetaData()

[...]

PR #139, opened 2016-10-24
Update index creation logic to use an explicit Index object

richardsmd:
Patch per #137 create Index object instead of issuing direct SQL.

All 125 tests pass per py.test - python 2.7/sa 1.0.4

### DIFF
--- a/geoalchemy2/__init__.py
+++ b/geoalchemy2/__init__.py
@@ -12,8 +12,16 @@ from .elements import (  # NOQA
 
 from . import functions  # NOQA
 
-from sqlalchemy import Table, event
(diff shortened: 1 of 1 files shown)


elemoine:
Unfortunately this does not work with PostGIS 1.x where geometry columns are created using `AddGeometryColumn`.

I actually had the same issue in my `spatial-index` branch, which I fixed by still using `spatial_index=True` with PostGIS < 2. See https://github.com/geoalchemy/geoalchemy2/compare/master...elemoine:spatial-index. That's not good, but I have no better idea for now.

elemoine:
Closing this. Please reopen if you know how to handle this.

PR #140, opened 2016-10-25
Set result_type when using operators

### LINKED ISSUE #93: Invalid SQL with load_only, order_by and limit
When running the example below, the following, invalid SQL query is generated:

``` sql
SELECT anon_1.venue_id              AS anon_1_venue_id, 
    St_asbinary(anon_1.venue_location)  AS anon_1_venue_location, 
    St_asbinary(anon_1.anon_2)      AS anon_1_anon_2, 
    label_1.id              AS label_1_id 
FROM    (
    SELECT venue.id AS venue_id, 
               venue.location AS venue_location, 
               venue.location <-> St_geomfromtext(:ST_GeomFromText_1, 
                  :ST_GeomFromText_2) AS anon_2 
    FROM   venue 
    ORDER  BY venue.location <-> St_geomfromtext(:ST_GeomFromText_1, 
                     :ST_GeomFromText_2) 
    LIMIT  :param_1
    ) AS anon_1 
LEFT OUTER JOIN (
    venue_to_label AS venue_to_label_1 
    JOIN label AS label_1 
    ON label_1.id = venue_to_label_1.label_id) 
ON anon_1.venue_id = venue_to_label_1.venue_id 
ORDER  BY anon_1.anon_2
```

The problem is that St_asbinary is applied to anon_1.anon_2. I'd expect the line either not to be generated or at least without the "St_asbinary". I'm pretty sure that this is the fault of GeoAlchemy2. Can anyone comment on that assumption?

_Any idea how to best resolve this issue? It seems pretty fundamental unfortunately. We're trying to use the code in a big project and any help is welcome!_

The (minimal) example below assumes a local PostgreSQL database "tmp" with GIS extension installed.

``` python
import unittest
from geoalchemy2 import WKTElement, Geometry
from sqlalchemy import Table, Column, Integer, ForeignKey
from sqlalchemy.ext.declarative import declarative_base
from sqlalchemy.orm import joinedload, relationship, load_only
from flask import Flask
from flask.ext.sqlalchemy import SQLAlchemy

app = Flask(__name__)
app.config['SQLALCHEMY_DATABASE_URI'] = (
    'postgres://postgres:password@localhost:5432/tmp')
db = SQLAlchemy(app)

Base = declarative_base()

# many (venue) <-> many (label) mapping table
venue_to_label = Table(
    'venue_to_label', db.metadata,
    Column('venue_id', Integer, ForeignKey('venue.id'), primary_key=True),
    Column('label_id', Integer, ForeignKey('label.id'), primary_key=True)
)


class Label(db.Model):
    __tablename__ = 'label'
    id = Column(Integer, primary_key=True, nullable=False)


class Venue(db.Model):
    id = Column(Integer, primary_key=True, nullable=False)
    labels = relationship(Label, secondary=venue_to_label)
    location = Column(Geometry(geometry_type="POINT"), nullable=False)

db.create_all()


class TestGeoAlchemy2Bug(unittest.TestCase):

    def test_geo_alchemy2_bug(self):
        point = WKTElement("POINT(0 0)")

        query = Venue.query
        query = query.options(joinedload(*['labels']).load_only(*['id']))
        query = query.order_by(Venue.location.distance_centroid(point))
        query = query.limit(10)

        print query
        print query.all()
```

### LINKED ISSUE #94: Using the <-> Operator
Hi

Im trying to generate this query in SQLAlchemy for performing fast distance ordering:

``` sql
SELECT ST_Transform(geom, 27700) <-> ST_Transform(ST_SetSRID(ST_MakePoint(-2.21335, 53.5628), 4326), 27700) AS distance
FROM foo
ORDER BY distance
LIMIT 10
```

So the code should be:

``` python
distance = (ST_Transform(Foo.geom, 27000).op('<->')(ST_Transform(func.ST_SetSRID(func.ST_MakePoint(-2.21335, 53.5628), 4326), 27700))).label('distance')
db.session.query(Foo).with_entities(distance).order_by(distance).limit(10)
```

However the generated SQL is:

``` sql
SELECT ST_AsBinary(ST_Transform(foo.geom, 27000) <-> ST_Transform(ST_SetSRID(ST_MakePoint(-2.21335, 53.5628), 4326), 27700)) AS distance
FROM foo 
ORDER BY distance
LIMIT 10
```

So it gets wrapped in `ST_AsBinary` which is incorrect and breaks the query.

Whats the best way to achieve the desired query?

Thanks,

Chris

elemoine:
Fixes #93
Fixes #94

### DIFF
--- a/geoalchemy2/comparator.py
+++ b/geoalchemy2/comparator.py
@@ -42,11 +42,30 @@ Reference
 ---------
 """
 
+from sqlalchemy import types as sqltypes
 from sqlalchemy.types import UserDefinedType
+from sqlalchemy.dialects.postgresql import DOUBLE_PRECISION
+from sqlalchemy.sql import operators
 try:
     from sqlalchemy.sql.functions import _FunctionGenerator
 except ImportError:  # SQLA < 0.9  # pragma: no cover
     from sqlalchemy.sql.expression import _FunctionGenerator
+
+
+INTERSECTS = operators.custom_op('&&')
+OVERLAPS_OR_TO_LEFT = operators.custom_op('&<')
+OVERLAPS_OR_TO_RIGHT = operators.custom_op('&>')
+OVERLAPS_OR_BELOW = operators.custom_op('&<|')
+TO_LEFT = operators.custom_op('<<')
+BELOW = operators.custom_op('<<|')
+TO_RIGHT = operators.custom_op('>>')
+CONTAINED = operators.custom_op('@')
+OVERLAPS_OR_ABOVE = operators.custom_op('|&>')
+ABOVE = operators.custom_op('|>>')
+CONTAINS = operators.custom_op('~')
+SAME = operators.custom_op('~=')
+DISTANCE_CENTROID = operators.custom_op('<->')
+DISTANCE_BOX = operators.custom_op('<#>')
 
 
 class BaseComparator(UserDefinedType.Comparator):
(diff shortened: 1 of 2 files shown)


elemoine:
> Do you maybe want to include the test case in #93 (comment)?

For some reason I did not get notified that you added a comment!

I just added new test cases for this. Will merge when the build passes.

simlu:
Great job, thank you very much!

PR #141, opened 2016-11-08
Add use_typmod option for AddGeometryColumn

tsauerwein:
[AddGeometryColumn](http://postgis.net/docs/AddGeometryColumn.html) has a parameter `use_typmod` which when set to `false` will create explicit constraints for dimension, SRID and geometry type instead of relying on the build-in types (for PostGIS 2.x).

This PR proposes to add a parameter `use_typmod` to `_GISType` which will be passed on to `AddGeometryColumn`.

### DIFF
--- a/geoalchemy2/__init__.py
+++ b/geoalchemy2/__init__.py
@@ -72,15 +72,18 @@ def _setup_ddl_event_listeners():
             for c in table.c:
                 # Add the managed Geometry columns with AddGeometryColumn()
                 if isinstance(c.type, Geometry) and c.type.management is True:
-                    stmt = select([
-                        func.AddGeometryColumn(
-                            table_schema,
-                            table.name,
-                            c.name,
-                            c.type.srid,
-                            c.type.geometry_type,
-                            c.type.dimension
-                        )])
+                    args = [
+                        table_schema,
+                        table.name,
+                        c.name,
+                        c.type.srid,
+                        c.type.geometry_type,
+                        c.type.dimension
+                    ]
+                    if c.type.use_typmod is not None:
+                        args.append(c.type.use_typmod)
+
+                    stmt = select([func.AddGeometryColumn(*args)])
                     stmt = stmt.execution_options(autocommit=True)
                     bind.execute(stmt)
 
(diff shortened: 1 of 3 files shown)


elemoine:
The code looks good to me.

Question: why would one want to use AddGeometryColumn with typmod set to False with PostGIS 2.0? I am just trying to understand the use-cases you address with this PR?

tsauerwein:
> Question: why would one want to use AddGeometryColumn with typmod set to False with PostGIS 2.0?

That is a good question. The use-case it indeed a bit special: We want to store geometries with 2D, 3D or 4D coordinates in the same geometry column (tracks with optionally elevation and timestamp). Using the PostGIS 2.* types, you would have to create a geometry column with 4 dimensions, and fill up the missing dimensions with `0`. Then you would need a 2nd column to remember the original dimension of your geometry.

With `typmod=false`, a trigger is created which checks the dimension. By removing the trigger manually, you can store geometries with whatever dimension in the same column.

Thanks for the review!

elemoine:
Ok, thanks for the explanation.

tsauerwein:
The merge of this PR [broke the build](https://travis-ci.org/geoalchemy/geoalchemy2/builds/176481187), because a new version of flake8 and pycodestyle were released on Monday, which apparently changed the linter rules. I will come up a with a fix.

elemoine:
@tsauerwein we can possibly publish a new version on pypi. Please tell me if you need that.

tsauerwein:
@elemoine A new release would be great, but there is no hurry from our side.

elemoine:
Also there's no problem giving you push access to pypi if necessary.

tsauerwein:
The package is uploaded automatically to PyPi during the Travis build (using your account)? I guess I will not even need access rights for PyPI. I can try to follow the procedure described here: https://github.com/geoalchemy/geoalchemy2/blob/master/RELEASE.rst

elemoine:
> The package is uploaded automatically to PyPi during the Travis build (using your account)? I guess I will not even need access rights for PyPI. I can try to follow the procedure described here: https://github.com/geoalchemy/geoalchemy2/blob/master/RELEASE.rst

Yep, I had forgotten about that.

PR #142, opened 2016-11-17
Adapt code style to new version of flake8/pycodestyle

tsauerwein:
A new version of flake8 and pycodestyle was released on Monday, which broke the Travis build. This PR fixes the linter issues.

Also I am proposing to pin down the versions of the development dependencies to have a reproducible build.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,8 +2,9 @@
 -e .
 
 # Additional requirements for running the testsuite and development
-flake8
-pytest
-pytest-cov
+pycodestyle==2.2.0
+flake8==3.2.0
+pytest==3.0.4
+pytest-cov==2.4.0
 
 Shapely>=1.3.0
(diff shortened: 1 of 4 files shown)


elemoine:
Please merge. Thanks.

requirements.txt
```diff
@@ -2,8 +2,9 @@
 -e .
 
 # Additional requirements for running the testsuite and development
-flake8
-pytest
-pytest-cov
+pycodestyle==2.2.0
+flake8==3.2.0
+pytest==3.0.4
+pytest-cov==2.4.0
```
elemoine: Actually, I am wondering about this. This could create conflicts when installing geoalchemy2 along other packages in a Python environment, no?

elemoine: Ok, I realize that `requirements.txt` is only used for development, so it's ok.

tsauerwein:
Thanks for the review, Eric!

PR #146, opened 2017-01-25
Add ST_AsTWKB function

JacobHayes:
Adds support for the ST_AsTWKB function as described here: http://postgis.net/docs/ST_AsEWKB.html


elemoine:
Thanks.

PR #152, opened 2017-05-18
Adds ability to pickle SpatialElements

katerina7479:
I need the ability to pickle WKBElements for some multiprocessing work I'm doing. Would really appreciate this functionality being included in your project!

### DIFF
--- a/geoalchemy2/elements.py
+++ b/geoalchemy2/elements.py
@@ -60,6 +60,20 @@ class _SpatialElement(object):
         func_ = functions._FunctionGenerator(expr=self)
         return getattr(func_, name)
 
+    # Allows for pickling
+    def __getstate__(self):
+
+        d = self.__dict__
+
+        return {'srid' : d['srid'], 'data' : str(self)}
+
+    def __setstate__(self, state):
+
+        args = [state['data']]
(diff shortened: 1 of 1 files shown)


elemoine:
It looks to me that `SpatialElement_`s are already pickable. Can you please provide a test-case demonstrating that they are not. Thank you.

katerina7479:
Well, I am using multiprocess, in conjunction with sqlalchemy with Geography locations. Specifically the WKBElement needs to be serializeable and pickleable, and it is not. I put the methods on the base class though. It's the whole system working together, and I can't really give you a short example, except to say it works on my fork, and not on your package.

elemoine:
Here is a script that works for me:

```py
import pickle
from geoalchemy2 import WKBElement


e = WKBElement(b'\x01\x02', srid=4326)
assert(e.desc == '0102')
assert(e.srid == 4326)

p = pickle.dumps(e)

n = pickle.loads(p)
assert(n.desc == e.desc)
assert(n.data == e.data)
assert(n.srid == e.srid)
```

elemoine:
@katerina7479 curious to know if that finally worked for you…

PR #153, opened 2017-05-19
Fix typo in elements code

elemoine:


elemoine:
Good point. I'll prepare a change log with a mention about this.

PR #162, opened 2017-08-01
Add checks to _GISType constructor

elemoine:
This PR add checks for the arguments passed to the `_GISType` constructor.

@tsauerwein, what do you think about that? I think it's a good idea to catch errors as early as possible, and warn the user when appropriate.


elemoine:
Thanks @tsauerwein. I added tests.

PR #166, opened 2017-09-22
Fix docstring for overlaps_or_above

dcere:
It is just a minor fix in a docstring: instead of comparing to see if it is to the right, we are comparing to see if it is above.


elemoine:
The build didn't pass but that's not related to this change. Probably related to Travis CI changes…

PR #170, opened 2018-01-10
Fix ST_LineLocatePoint return type

fredj:
The function returns a number, not a `types.Geometry`.
See https://postgis.net/docs/ST_LineLocatePoint.html


elemoine:
Thanks @fredj

fredj:
Thanks a lot for the fast merge !

PR #175, opened 2018-04-07
Add support for function ST_Azimuth.

simlmx:


geoalchemy2/functions.py
```diff
@@ -186,6 +186,11 @@ def __init__(self, *args, **kwargs):
      'For ``geometry`` type area is in SRID units. For ``geography`` area is '
      'in square meters.'),
 
+    ('ST_Azimuth', None,
+     'ST_Azimuth - Returns the angle in radians from the horizontal of the '
```
elemoine: You can remove "ST_Azimuth -" entirely, and starts the sentence with "Returns the angle…". This is to be consistent with the other functions.

simlmx: Right, I should've noticed that! Thanks.

simlmx: Done in 5143c5b6197f3dcb4bd754d0e169beff63f7b6ca.

PR #177, opened 2018-05-25
fix  pickle can't handle.

hifans:
fix the PostGIS returns a Buffer object for geometry columns, which pickle can't handle.

objects = session.query(MyGeoObject).all()
pickle.dumps(objects)

You'll get :

TypeError: can't pickle buffer objects

The problem is that PostGIS returns a Buffer object for geometry columns, which pickle can't handle.


elemoine:
This breaks the tests.

elemoine:
Superseded by https://github.com/geoalchemy/geoalchemy2/pull/182.

PR #185, opened 2018-07-16
Fix to_shape with SpatiaLite

### LINKED ISSUE #184: to_shape not working on spatialite
Hello,

After some tests on latest version, for retrieving a shapely geom I have to use : 
- Spatialite base : `shapely.wkb.loads(row.geom.data,hex=True)` (on postgis : _attributeError: 'memoryview' object has no attribute 'encode'_)
- Postgis base : `to_shape(row.geom)` (on spatialite : _TypeError: string argument without an encoding_)

I don't really know why, but maybe it can help, so I put it here.

Regards,

elemoine:
Fixes #184.

### DIFF
--- a/geoalchemy2/shape.py
+++ b/geoalchemy2/shape.py
@@ -11,7 +11,7 @@ import shapely.wkb
 import shapely.wkt
 
 from .elements import WKBElement, WKTElement
-from .compat import buffer, bytes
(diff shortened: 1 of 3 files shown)


elemoine:
All the tests pass. Will merge this in the coming days. Happy to get reviews!

Issue #157, opened 2017-07-26
clarify 3d geometry type support


tomkralidis:
Not sure if I'm looking in the right place, but the [docs](https://geoalchemy-2.readthedocs.io/en/latest/types.html#geoalchemy2.types._GISType) do not specify support for 3d geometries, even though the following seems to work:

```python
 location = Column(Geometry('POINTZ', dimension=3, srid=4326))
```

elemoine:
Right. I think the doc needs to be extended.

adrien-berchet:
Fixed in `0.6.3`

Issue #67, opened 2013-11-22
dimension support


iocast:
It tried to define a geometry attribute using 3 dimension. But when I create the tables, it always creates a 2 dimension geometry. Could that be a bug?

```
class Boundary(Base):
    __tablename__ = 'boundaries'
    ...
    geom = Column(Geometry(geometry_type='MULTIPOLYGON', dimension=3, spatial_index=True), nullable=False)
    …
```

and

```
Base.metadata.create_all(create_engine(self.uri, echo=False), checkfirst=True)
```

Juanlu001:
Related question on Stack Overflow:

https://stackoverflow.com/q/37347770/554319

Potential duplicate of this issue:

https://github.com/geoalchemy/geoalchemy2/issues/157

Possible actions:

* Document workaround (using `*Z` geometry types)
* Raise warnings or errors when using the `dimension` parameter with 2D geometry types

Would the maintainers accept a pull request fixing this?

elemoine:
@Juanlu001 sure!

damb:
By chance I stumbled over this issue.

Opened a PR (#195). I went for raising `exc.ArgumentError` in case of incompatible `geometry_type` and `dimension` parameters (as mentioned by @Juanlu001).

PR #195, opened 2018-09-07
Dimension validation

damb:
Raise an `exc.ArgumentError` in case of incompatible `geometry_type` and `dimension` parameters. Avoid the issues mentioned in #67 and #157.


elemoine:
Looks great!

PR #204, opened 2019-01-08
 Define compiled versions of functions for SpatiaLite 

elemoine:
The main purpose of this PR is to remove the `use_st_prefix` argument to the `Geometry` and `Geography` types and, instead, use `@compiles('sqlite')` for SpatiaLite functions that don't have the `ST_` prefix. As a nice bonus this removes one of the SpatiaLite-specific caveats that we documented in the SpatiaLite tutorial.

Thanks @TomGoBravo for fostering this work.


TomGoBravo:
+1
This code and tests look good to me. I was hoping to get some time to compare to the odd pattern of ST_ and no ST_ functions in SpatiaLite but don't want to leave you hanging any longer.

PR #209, opened 2019-02-27
Unpack SRID from WKB struct when it is not provided as WKBElement arg

### LINKED ISSUE #208: SRID lost when using ST_Tranform function
Hello,

I am unable to retrieve SRID on WKBElement, when using ST_Tranform function in select. The SRID is set to `-1` (default value). It does not sound like an expected behaviour.

Environment info:
- postgres 9.6
- postgis 2.3
- python 2.7.12
- psycopg2 2.7.7
- sqlalchemy 1.2.18
- geoalchemy2 0.5.0

You can find below the steps to reproduce the issue.

```python
from geoalchemy2 import Geometry, WKBElement
from geoalchemy2.shape import from_shape, to_shape
from shapely.geometry import Point
from sqlalchemy import Column, create_engine, func, insert, Integer, MetaData, select, Table

conn = create_engine('postgresql://username:password@host:5432/database').connect()

obj = Table(
    'obj', MetaData(),
    Column('id', Integer, primary_key=True, nullable=False),
    Column('geom', Geometry(srid=4326), nullable=True),
)
obj.create(conn)
conn.execute(insert(obj).values(id=1, geom=from_shape(Point(1, 2), srid=4326)))

# test 1: no transformation -> OK
geom = conn.scalar(select([obj.c.geom]).where(obj.c.id == 1))
assert isinstance(geom, WKBElement)
assert geom.srid == 4326
assert to_shape(geom).wkt == 'POINT (1 2)'

# test 2: transform to 2154 -> OK but SRID lost
geom = conn.scalar(select([func.ST_Transform(obj.c.geom, 2154)]).where(obj.c.id == 1))
assert isinstance(geom, WKBElement)
assert geom.srid == -1
assert to_shape(geom).wkt == 'POINT (409711.8296099951 1197127.908805346)'

# test 3: try to preserve SRID
geom = conn.scalar(select([func.ST_SetSRID(func.ST_Transform(obj.c.geom, 2154), 2154)]).where(obj.c.id == 1))
assert isinstance(geom, str)
assert geom == '01010000206A08000055488551BF0119419877A7E847443241'

# test 4: try to preserve WKBElement
geom = conn.scalar(select([func.ST_AsEWKB(func.ST_SetSRID(func.ST_Transform(obj.c.geom, 2154), 2154))]).where(obj.c.id == 1))
assert isinstance(geom, buffer)
assert str(buffer) == '\x01\x01\x00\x00 j\x08\x00\x00UH\x85Q\xbf\x01\x19A\x98w\xa7\xe8GD2A'
```

Thank you! :smiley:

SergeBouchut:
Now geometry result processor preserves query SRID.

```python
geom = conn.scalar(select([func.ST_Transform(func.ST_GeomFromEWKT('SRID=4326;POINT(1 1)'), 2154, type=Geometry(srid=2154))])) 
assert geom.srid == 2154  # previously -1
```

This is made possible by unpacking SRID from WKB struct, when instanciate `WKBElement` without `srid` argument.

```python
geom = WKBElement(b'\x01\x01\x00\x00\x00\xcf\x02\x14\rO\xc8\x18A\x08`\xe4Jw\r0A', extended=True)
assert geom.srid == 2154  # previously -1
```

Note:
- I have swapped  `TestWKBElement` with `TestExtendedWKBElement` test class names.
- Also I set a real WKB binary for `TestExtendedWKBElement` test class: the dummy previous one could not be unpacked.
- It seems that some test cases in `test_functional_spatialite` module, initialize `WKBElement` with hex value (instead of bin), so I have to support this too.

Fixes #208

### DIFF
--- a/geoalchemy2/elements.py
+++ b/geoalchemy2/elements.py
@@ -1,4 +1,5 @@
 import binascii
+import struct
 from geoalchemy2.compat import PY3, str as str_
 
 try:
@@ -131,10 +132,39 @@ class WKBElement(_SpatialElement):
 
     Note: you can create ``WKBElement`` objects from Shapely geometries
     using the :func:`geoalchemy2.shape.from_shape` function.
+
+    Extended WKB automatically parse unknown SRID.
+    Warning: unicode data are assumed to be hexadecimal value.
+
+    WKB struct {
+        byte    byteOrder;
+        uint32  wkbType;
+        uint32  SRID;
+        struct  geometry;
+    }
+
+    byteOrder enum {
+        WKB_XDR = 0,  // Most Significant Byte First
+        WKB_NDR = 1,  // Least Significant Byte First
+    }
+
     """
 
     geom_from = 'ST_GeomFromWKB'
     geom_from_extended_version = 'ST_GeomFromEWKB'
+
+    def __init__(self, data, srid=-1, extended=False):
+        if extended and srid == -1:  # unpack srid from WKB struct
+            if isinstance(data, str_):  # unicode data assumed to be hex val
+                header = binascii.unhexlify(data[:18])
+            else:
+                header = data[:9]
+            if not PY3:
+                header = bytearray(header)
+            byte_order, srid = header[0], header[5:]
+            srid = struct.unpack('<I' if byte_order else '>I', srid)[0]
+
+        _SpatialElement.__init__(self, data, srid, extended)
 
     @property
     def desc(self):
--- a/tests/test_elements.py
+++ b/tests/test_elements.py
@@ -1,12 +1,13 @@
 import re
 import pytest
 
+from shapely import wkb
 from sqlalchemy import Table, MetaData, Column, String, func
 from geoalchemy2.types import Geometry
 from geoalchemy2.elements import (
     WKTElement, WKBElement, RasterElement, CompositeElement
 )
-from geoalchemy2.compat import str as str_
+from geoalchemy2.compat import buffer as buffer_, bytes as bytes_, str as str_
 
 
 @pytest.fixture
@@ -128,39 +129,47 @@ class TestExtendedWKTElementFunction():
         }
 
 
-class TestWKBElement():
+class TestExtendedWKBElement():
+
+    # _bin/_hex computed by following query:
+    # SELECT ST_GeomFromEWKT('SRID=3;POINT(1 2)');
+    _bin = buffer_(b'\x01\x01\x00\x00 \x03\x00\x00\x00\x00\x00\x00'
+                   b'\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\x00@')
+    _hex = str_('010100002003000000000000000000f03f0000000000000040')
+    _srid = 3  # expected srid
+    _wkt = 'POINT (1 2)'  # expected wkt
 
     def test_desc(self):
-        e = WKBElement(b'\x01\x02', extended=True)
-        assert e.desc == '0102'
+        e = WKBElement(self._bin, extended=True)
+        assert e.desc == self._hex
 
     def test_desc_str(self):
-        e = WKBElement(str_('0102'))
-        assert e.desc == str_('0102')
+        e = WKBElement(self._hex)
+        assert e.desc == self._hex
 
     def test_function_call(self):
-        e = WKBElement(b'\x01\x02', extended=True)
+        e = WKBElement(self._bin, extended=True)
         f = e.ST_Buffer(2)
         eq_sql(f, 'ST_Buffer('
                'ST_GeomFromEWKB(:ST_GeomFromEWKB_1), '
                ':ST_Buffer_1)')
         assert f.compile().params == {
             u'ST_Buffer_1': 2,
-            u'ST_GeomFromEWKB_1': b'\x01\x02',
+            u'ST_GeomFromEWKB_1': self._bin,
         }
 
     def test_function_str(self):
-        e = WKBElement(b'\x01\x02', extended=True)
+        e = WKBElement(self._bin, extended=True)
         assert isinstance(str(e), str)
 
     def test_pickle_unpickle(self):
         import pickle
-        e = WKBElement(b'\x01\x02', srid=3, extended=True)
+        e = WKBElement(self._bin, srid=self._srid, extended=True)
         pickled = pickle.dumps(e)
         unpickled = pickle.loads(pickled)
-        assert unpickled.srid == 3
+        assert unpickled.srid == self._srid
         assert unpickled.extended is True
-        assert unpickled.data == b'\x01\x02'
+        assert unpickled.data == bytes_(self._bin)
         assert unpickled.name == 'ST_GeomFromEWKB'
         f = unpickled.ST_Buffer(2)
         eq_sql(f, 'ST_Buffer('
@@ -168,11 +177,25 @@ class TestWKBElement():
                ':ST_Buffer_1)')
         assert f.compile().params == {
             u'ST_Buffer_1': 2,
-            u'ST_GeomFromEWKB_1': b'\x01\x02',
+            u'ST_GeomFromEWKB_1': bytes_(self._bin),
         }
 
+    def test_unpack_srid_from_bin(self):
+        """
+        Unpack SRID from WKB struct (when it is not provided as arg)
+        to ensure geometry result processor preserves query SRID.
+        """
+        e = WKBElement(self._bin, extended=True)
+        assert e.srid == self._srid
+        assert wkb.loads(bytes_(e.data)).wkt == self._wkt
 
-class TestExtendedWKBElement():
+    def test_unpack_srid_from_hex(self):
+        e = WKBElement(self._hex, extended=True)
+        assert e.srid == self._srid
+        assert wkb.loads(e.data, hex=True).wkt == self._wkt
+
+
+class TestWKBElement():
 
     def test_desc(self):
         e = WKBElement(b'\x01\x02')


geoalchemy2/elements.py
```diff
@@ -136,6 +137,27 @@ class WKBElement(_SpatialElement):
     geom_from = 'ST_GeomFromWKB'
     geom_from_extended_version = 'ST_GeomFromEWKB'
 
+    WKB_XDR = 0  # Most Significant Byte First
```
loicgasser: not used?

geoalchemy2/elements.py
```diff
@@ -136,6 +137,27 @@ class WKBElement(_SpatialElement):
...
+    WKB_NDR = 1  # Least Significant Byte First
+
+    def __init__(self, data, srid=-1, extended=False):
+        if extended and srid == -1:
+            # unpack srid from WKB struct
+
+            try:
+                # support hex input
+                data = binascii.unhexlify(data)
```
loicgasser: maybe use `self._data_from_desc` ?

geoalchemy2/elements.py
```diff
@@ -136,6 +137,27 @@ class WKBElement(_SpatialElement):
...
+            except (binascii.Error, TypeError):
+                pass
+            byte_order, srid = data[0], data[5:9]
+
+            if not isinstance(byte_order, int):  # py2
```
loicgasser: `if not PY3:` ?

geoalchemy2/elements.py
sbouchut: yep, should I move it in docstring? (to keep the postgis enum as documentation purpose)

SergeBouchut: moved to docstring

geoalchemy2/elements.py
SergeBouchut: finally I used slicing to not damage perf with big geometries

geoalchemy2/elements.py
SergeBouchut: ok

geoalchemy2/elements.py
```diff
@@ -131,11 +132,35 @@ class WKBElement(_SpatialElement):
...
     geom_from = 'ST_GeomFromWKB'
     geom_from_extended_version = 'ST_GeomFromEWKB'
 
+    def __init__(self, data, srid=-1, extended=False):
+        if extended and srid == -1:
+            # unpack srid from WKB struct
+
+            try:
+                # support hex input
+                data = binascii.unhexlify(data[:18])
```
adrien-berchet-forcity: @sbouchut I just realized that we should create a specific variable instead of erasing `data` because it is then passed to `_SpatialElement.__init__()` so in the end the WKBElement.data will be ill-formed. Note that the tests do not point this unexpected behaviour, maybe we should add a test for that.

SergeBouchut: agree, thanks for the feedback

geoalchemy2/elements.py
```diff
@@ -131,11 +132,41 @@ class WKBElement(_SpatialElement):
...
+
     """
 
     geom_from = 'ST_GeomFromWKB'
     geom_from_extended_version = 'ST_GeomFromEWKB'
 
+    def __init__(self, data, srid=-1, extended=False):
+        if extended and srid == -1:  # unpack srid from WKB struct
+            header = data[:9]
+            try:  # support hex input
```
elemoine: Can you please remind me when we can get "hex input"? Is "hex input" something we already support? Or is it a new feature you are adding?

elemoine:
I just added a question, but I like the general idea of this patch! Thanks.

geoalchemy2/elements.py
elemoine: Oh I guess we guess an hex string when using SpatiaLite.

Why not be consistent with the desc function and do this:

```py
if isinstance(data, str_):
    header = binascii.unhexlify(data[:18])
else:
    header = data[:9]
```

?

geoalchemy2/elements.py
```diff
@@ -131,11 +132,41 @@ class WKBElement(_SpatialElement):
...
+                header = binascii.unhexlify(data[:18])
+            except (binascii.Error, TypeError):
+                pass
+
+            byte_order, srid = header[0], header[5:]
+            if not PY3:
+                byte_order = int(binascii.hexlify(byte_order), 16)
```
elemoine: I'd like to see a comment about why this extra step is necessary with Python 2

elemoine:
This looks great. A few things and we can merge this.

geoalchemy2/elements.py
```diff
@@ -131,11 +132,41 @@ class WKBElement(_SpatialElement):
...
+        WKB_NDR = 1,  // Least Significant Byte First
+    }
+
     """
 
     geom_from = 'ST_GeomFromWKB'
     geom_from_extended_version = 'ST_GeomFromEWKB'
 
+    def __init__(self, data, srid=-1, extended=False):
+        if extended and srid == -1:  # unpack srid from WKB struct
```
adrien-berchet-forcity: I was wondering if the `srid == -1` statement is relevent. Maybe we could remove it and raise an error if the given srid is not `-1` is not consistent with the one contained in the data?

elemoine: I was not sure either. I also like the idea of avoiding the data unpacking when the srid is specified in the Geometry type.

For example, if I am always transforming to, let's say, EPSG:3857, I can write my own function class like this:

```py
class ST_Transform3857(GenericFunction):
    type = Geometry(srid=3857)
    name = 'ST_Transform'
    identifier = 'ST_Transform3857'
```

Or I can set the type at query time:

```py
s = select([func.ST_Transform(t.c.geom, 4326), 3857, _type=Geometry(srid=3857))])
```

So, with the way @SergeBouchut wrote the code we avoid the data unpacking in the above two cases.

adrien-berchet-forcity: Ok I see. It is clearly more efficient in those cases but the user is responsible to ensure that the 2 srid are consistent. Maybe we could just add a word in the doc to make it clear?

elemoine: Agreed.

geoalchemy2/elements.py
SergeBouchut: It is not clear for me if hex input are supported (I find no mention in doc). You are right, I added it because of SpatiaLite tests.
I am not sure that checking data is instance of string is exact, because it seems to me that binary input could also be a string (not only hex input).
Maybe we need to change the desc function?

elemoine: > It is not clear for me if hex input are supported (I find no mention in doc). You are right, I added it because of SpatiaLite tests.

With SQLite/SpatiaLite we do get an hex string.

> I am not sure that checking data is instance of string is exact, because it seems to me that binary input could also be a string (not only hex input).

In the case of SpatiaLite we get a string, in the case of Postgis we get a memoryview (or a buffer in the case of Python 2).

So I think that the `isinstance(data, str_)` test that we use in `desc` is correct. I'd vote for consistency and using the same test in both `desc` and `__init__`.

geoalchemy2/elements.py
SergeBouchut:
```python
# bytes string slicing not automatically produces int output in py2
# py2: b'\x01\x00'[0] -> '\x01'
# py3: b'\x01\x00'[0] -> 1
```

SergeBouchut: maybe this is more elegant (and clear):
```python
if not PY3:
    header = bytearray(header)
byte_order, srid = header[0], header[5:]
```

geoalchemy2/elements.py
SergeBouchut: In the case of SpatiaLite and Python do we always get a **unicode** string?
Otherwise I also have to check if input is instance of `bytes`.

geoalchemy2/elements.py
elemoine: Perfect for the comment. Please add it to the code. And +1 for using `bytearray` for Python 2 if that works as expected.

geoalchemy2/elements.py
elemoine: In the case of SpatiaLite we get a unicode with Python 2 and a str with Python 3. This is why I am suggesting to use `isinstance(data, str_)`.

elemoine:
It sounds like we're close to the final version. Thank you @SergeBouchut for the continued effort on this. Great addition!

geoalchemy2/elements.py
SergeBouchut: ok :+1:

tests/test_elements.py
```diff
@@ -128,51 +129,73 @@ def test_ST_Equal_Column_WKTElement(self, geometry_table):
...
         import pickle
-        e = WKBElement(b'\x01\x02', srid=3, extended=True)
+        e = WKBElement(self._bin, srid=self._srid, extended=True)
         pickled = pickle.dumps(e)
         unpickled = pickle.loads(pickled)
-        assert unpickled.srid == 3
+        assert unpickled.srid == self._srid
         assert unpickled.extended is True
-        assert unpickled.data == b'\x01\x02'
+        assert unpickled.data == bytes_(self._bin)
```
SergeBouchut: I am not sure to understand why unpickled data ara `bytes` and not `buffer`...

tests/test_elements.py
```diff
@@ -128,51 +129,73 @@ def test_ST_Equal_Column_WKTElement(self, geometry_table):
...
+    _bin = buffer_(b'\x01\x01\x00\x00 \x03\x00\x00\x00\x00\x00\x00'
+                   b'\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\x00@')
+    _hex = str_('010100002003000000000000000000f03f0000000000000040')
+    _srid = 3  # expected srid
+    _wkt = 'POINT (1 2)'  # expected wkt
 
     def test_desc(self):
-        e = WKBElement(b'\x01\x02', extended=True)
-        assert e.desc == '0102'
+        e = WKBElement(self._bin, extended=True)
```
SergeBouchut: Previously test data where `bytes`, but we assume Postgis only returns `buffer`.

geoalchemy2/elements.py
SergeBouchut: done

tests/test_elements.py
elemoine: Right.

tests/test_elements.py
elemoine: I agree that relying on `buffer_` is better here.

geoalchemy2/elements.py
```diff
@@ -131,11 +132,38 @@ class WKBElement(_SpatialElement):
...
+            header = (binascii.unhexlify(data[:18]) if isinstance(data, str_)
+                      else data[:9])  # unicode data assumed to be hex val
```
elemoine: I am picky but I find this form more readable:

```py
if isinstance(data, str_):
    header = binascii.unhexlify(data[:18])
else:
    header = data[:9]
```

Can you please change it? Otherwise I'll do it myself once this is merged :)

SergeBouchut: Of course, no problem.
I chose this way because it is safer to ensure `header` is set before used :wink: 
Otherwise, an alternative is:
```python
header = data[:9]
if isinstance(data, str_):
    header = binascii.unhexlify(data[:18])
```
but the first slicing is useless, in the case of unicode `data`.

elemoine: Relying on `else` is safe too. It is garanteed that `header` will be defined after the `if/else` block. Let's change this and merge! ;-)

elemoine:
And please squash your commits into one commit, whose head message starts with a capital letter. Here the commit message could for example be "Change WKBElement to read SRID from the EWKB string". Thanks.

geoalchemy2/elements.py
SergeBouchut: :+1:

elemoine:
I'll see if I can add some docs as suggested by Adrien.

PR #210, opened 2019-03-02
Improve docs related to EWKB and SRID

elemoine:
This PR follows up on #209.

Ping @SergeBouchut and @adrien-berchet-forcity. Also, please tell me if you need a new geoalchemy2 release on pypi.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -4,7 +4,7 @@ Release
 This file provides the steps for releasing a new version of GeoAlchemy 2.
 
 Add a new section to CHANGES.txt, change the version number in ``setup.py`` and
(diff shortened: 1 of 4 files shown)


adrien-berchet-forcity:
The doc is fine for me.
A new release would be cool for us since we delayed a feature in our internal tools because of this. But shouldn't we make the `WKTElement()` behaviour consistent with the new `WKBElement()`, i.e. extracting the SRID from EWKT when `srid=-1`? It should not be hard so I can make a PR for this today.

adrien-berchet-forcity:
Proposal: #211

elemoine:
Not sure about WKTElement. I'll have a closer look when I can.

PR #211, opened 2019-03-04
Change WKTElement to read SRID from the EWKT string.

adrien-berchet-forcity:
Make WKTElement behaviour consistent with the WKBElement one introduced in #209.

### DIFF
--- a/geoalchemy2/elements.py
+++ b/geoalchemy2/elements.py
@@ -1,6 +1,5 @@
 import binascii
 import struct
-from geoalchemy2.compat import PY3, str as str_
 
 try:
     from sqlalchemy.sql import functions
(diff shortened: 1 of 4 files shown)


elemoine:
Not sure I agree with that change but I'll have a closer look.

elemoine:
I was hesitant because I think there may be a problem [here](https://github.com/geoalchemy/geoalchemy2/blob/master/geoalchemy2/types.py#L146), where that line is not compatible with extended WKTElement objects. But this problem exists independently from your change actually.

elemoine:
We need to add more tests to TestGeometry.

adrien-berchet-forcity:
Ok I see your point. I updated a test and proposed a simple fix in bind_processor() to point the issue. But it probably could be done in a better way.

But maybe you prefer keeping the old behavior?

elemoine:
I just want to make sure that we're not breaking backward compatibility. It looks like we're not. Do you agree?

elemoine:
I made changes directly to your branch. Happy to merge it if you agree. Thanks!

adrien-berchet-forcity:
I also think we don't break anything with this new version. Thank you for your last updates.

Issue #215, opened 2019-03-21
PostGIS functions are case sensitive


adrien-berchet:
Hi,
We just discovered that the PostGIS functions are case sensitive when called with the following syntax:
```python
from sqlalchemy import select, func
conn.execute(select([func.ST_Transform(table.c.geometry, 2154)])).fetchall()
conn.execute(select([func.st_transform(table.c.geometry, 2154)])).fetchall()
```
The first one returns a `WKBElement` as expected while the second one returns the EWKB as string. On the contrary, `select([city_objects.c.geometry.ST_Transform(2154)])` returns the expected `WKBElement` while `select([city_objects.c.geometry.st_transform(2154)])` raises the following error: `AttributeError: Neither 'Column' object nor 'Comparator' object has an attribute 'st_transform'`.

I think that these behaviours are quite confusing and could be improved. I see 2 solutions: either raising the same Attribute error when calling `func.st_transform` (or any lowercase function), or make function calls case insensitive.

I think I prefer the second option. What do you think about it?

elemoine:
Excellent work on https://github.com/sqlalchemy/sqlalchemy/commit/702c96b2ee94a394ab1b4d528e3090c4d3c42b91 @adrien-berchet! Does this mean there's nothing to do on the GeoA side?

adrien-berchet:
Thanks :blush: 
I will do some tests but I think that sqlalchemy/sqlalchemy@702c96b fixes almost everything for this issue.

adrien-berchet:
Fixed in #216

PR #216, opened 2019-03-21
Lowercase PostGIS functions

adrien-berchet:
Proposal for #215 (I chose to allow lowercase function in this proposal, which might not be what you want)

I am not sure it is the best way to do it but it seems to work as far as I can see.

### DIFF
--- a/geoalchemy2/comparator.py
+++ b/geoalchemy2/comparator.py
@@ -86,7 +86,7 @@ class BaseComparator(UserDefinedType.Comparator):
         # This is not to mess up with SQLAlchemy's use of
         # hasattr/getattr on Column objects.
 
-        if not name.startswith('ST_'):
+        if not name.lower().startswith('st_'):
             raise AttributeError
 
         # We create our own _FunctionGenerator here, and use it in place of
--- a/tests/test_functional.py
+++ b/tests/test_functional.py
@@ -1,3 +1,4 @@
+from pkg_resources import parse_version
 import pytest
 
 try:
@@ -8,6 +9,7 @@ else:
     compat.register()
     del compat
 
+from sqlalchemy import __version__ as SA_VERSION
 from sqlalchemy import create_engine, Table, MetaData, Column, Integer, bindparam
 from sqlalchemy.orm import sessionmaker
 from sqlalchemy.ext.declarative import declarative_base
@@ -488,6 +490,40 @@ class TestCallFunction():
             .filter(Poi.geog.ST_Distance(
                 type_coerce('POINT(5 45)', Geography)) < 1000).one()
         assert poi.id == poi_id
+
+    @pytest.mark.skipif(
+        parse_version(SA_VERSION) < parse_version("1.3.4"),
+        reason='Case-insensitivity is only available for sqlalchemy>=1.3.4')
+    def test_comparator_case_insensitivity(self):
+        lake_id = self._create_one_lake()
+
+        s = select([func.ST_Buffer(Lake.__table__.c.geom, 2)])
+        r1 = session.execute(s).scalar()
+        assert isinstance(r1, WKBElement)
+
+        lake = session.query(Lake).get(lake_id)
+
+        r2 = session.execute(lake.geom.ST_Buffer(2)).scalar()
+        assert isinstance(r2, WKBElement)
+
+        r3 = session.execute(lake.geom.st_buffer(2)).scalar()
+        assert isinstance(r3, WKBElement)
+
+        r4 = session.execute(lake.geom.St_BuFfEr(2)).scalar()
+        assert isinstance(r4, WKBElement)
+
+        r5 = session.query(Lake.geom.ST_Buffer(2)).scalar()
+        assert isinstance(r5, WKBElement)
+
+        r6 = session.query(Lake.geom.st_buffer(2)).scalar()
+        assert isinstance(r6, WKBElement)
+
+        r7 = session.query(Lake.geom.St_BuFfEr(2)).scalar()
+        assert isinstance(r7, WKBElement)
+
+        assert (
+            r1.data == r2.data == r3.data == r4.data == r5.data == r6.data
+            == r7.data)
 
 
 class TestReflection():
--- a/tests/test_functional_spatialite.py
+++ b/tests/test_functional_spatialite.py
@@ -1,8 +1,10 @@
+from pkg_resources import parse_version
 import os
 import pytest
 import platform
 import json
 
+from sqlalchemy import __version__ as SA_VERSION
 from sqlalchemy import create_engine, MetaData, Column, Integer
 from sqlalchemy.orm import sessionmaker
 from sqlalchemy.ext.declarative import declarative_base
@@ -252,3 +254,37 @@ class TestCallFunction():
 
         r = session.query(Lake.geom.ST_AsGeoJSON()).scalar()
         _test(r)
+
+    @pytest.mark.skipif(
+        parse_version(SA_VERSION) < parse_version("1.3.4"),
+        reason='Case-insensitivity is only available for sqlalchemy>=1.3.4')
+    def test_comparator_case_insensitivity(self):
+        lake_id = self._create_one_lake()
+
+        s = select([func.ST_Buffer(Lake.__table__.c.geom, 2)])
+        r1 = session.execute(s).scalar()
+        assert isinstance(r1, WKBElement)
+
+        lake = session.query(Lake).get(lake_id)
+
+        r2 = session.execute(lake.geom.ST_Buffer(2)).scalar()
+        assert isinstance(r2, WKBElement)
+
+        r3 = session.execute(lake.geom.st_buffer(2)).scalar()
+        assert isinstance(r3, WKBElement)
+
+        r4 = session.execute(lake.geom.St_BuFfEr(2)).scalar()
+        assert isinstance(r4, WKBElement)
+
+        r5 = session.query(Lake.geom.ST_Buffer(2)).scalar()
+        assert isinstance(r5, WKBElement)
+
+        r6 = session.query(Lake.geom.st_buffer(2)).scalar()
+        assert isinstance(r6, WKBElement)
+
+        r7 = session.query(Lake.geom.St_BuFfEr(2)).scalar()
+        assert isinstance(r7, WKBElement)
+
+        assert (
+            r1.data == r2.data == r3.data == r4.data == r5.data == r6.data
+            == r7.data)


elemoine:
I'll have a look. Thanks for the patch.

geoalchemy2/functions.py
```diff
@@ -392,9 +392,14 @@ def __init__(self, *args, **kwargs):
...
 
+
 # Iterate through _FUNCTION and create GenericFunction classes dynamically
-for name, type_, doc in _FUNCTIONS:
-    attributes = {'name': name}
+def add_postgis_function(name, type_, doc, lowercase=False):
+    if lowercase:
+        lowercase_name = name.lower()
+    else:
+        lowercase_name = name
```
elemoine: I would not introduce a new local var. I'd use this:

```py
if lowecase:
    name = name.lower()
```

geoalchemy2/functions.py
```diff
@@ -392,9 +392,14 @@ def __init__(self, *args, **kwargs):
       'in then reads it from metadata of raster.', 'RT_ST_Value')),
 ]
 
+
 # Iterate through _FUNCTION and create GenericFunction classes dynamically
-for name, type_, doc in _FUNCTIONS:
-    attributes = {'name': name}
+def add_postgis_function(name, type_, doc, lowercase=False):
```
elemoine: It's not only PostGIS so I'd use another function name.

elemoine:
I like it. I just added a few comments on the code.

Also I wonder if it's worth adding so many tests. Can't we add just a few tests to test_functions.py?

Thanks

adrien-berchet-forcity:
Thanks for your comments, I updated according to them.

But I am not fully satisfied with this fix because it only allows to choose between lowercase and usual case, but a typo like `func.St_TrAnSfOrM()` still returns a string silently instead of a WKBElement.  Do you think we could fix this behaviour for any case?

I agree for the tests, I will rework them when I get satisfied with the fix.

elemoine:
>  but a typo like `func.St_TrAnSfOrM()` still returns a string silently instead of a WKBElement.

Oh I missed that. You're right, it doesn't feel correct to fix the problem for st_transform and not for St_TraNsForm. So I am not satisfied either.

adrien-berchet-forcity:
I found a tricky solution but it is quite unorthodox: I replace the dictionary used in `sqlalchemy.sql.functions._registry` by a new data structure that makes the calls to ST_* functions case insensitive. But this trick might introduce conflicts with some other packages if they use the same trick so I think we should use a specific func package like `func.geo.ST_*` to be sure that this trick is restricted to `geoalchemy2` functions (but it would be a breaking change, while this version is not).

elemoine:
> I found a tricky solution but it is quite unorthodox: I replace the dictionary used in `sqlalchemy.sql.functions._registry` by a new data structure that makes the calls to ST_* functions case insensitive.

Sounds too hackish to me.

> so I think we should use a specific func package like `func.geo.ST_*` to be sure that this trick is restricted to `geoalchemy2` functions (but it would be a breaking change, while this version is not).

I would not like to introduce such a breaking change.

My favorite option for the moment is Status quo :(

adrien-berchet-forcity:
I agree that `geoalchemy2` should not dynamically hack `sqlalchemy` to make it fit its needs. I think that this feature (allowing case-insensitivity to functions) should be added in `sqlalchemy.sql.functions` and then used in `geoalchemy2`. If you agree with this, I will make a proposal to `sqlalchemy` and then another one here to use this new feature.

elemoine:
I was thinking of the same.

geoalchemy2/comparator.py
```diff
@@ -86,7 +86,7 @@ def __getattr__(self, name):
         # This is not to mess up with SQLAlchemy's use of
         # hasattr/getattr on Column objects.
 
-        if not name.startswith('ST_'):
+        if not name.lower().startswith('st_'):
```
elemoine: With https://github.com/sqlalchemy/sqlalchemy/commit/702c96b2ee94a394ab1b4d528e3090c4d3c42b91 this change may still be useful. @adrien-berchet interested to have your advice about all this.

adrien-berchet: Indeed, I think this is the only change we need in geoalchemy2. I will test that.

geoalchemy2/comparator.py
```diff
@@ -52,6 +53,8 @@
     from sqlalchemy.sql.expression import _FunctionGenerator
 
 
+SA_VERSION = [int(i) for i in SA_VERSION.split('.')[:3]]
```
elemoine: Wouldn't that break with beta versions? For example: https://pypi.org/project/SQLAlchemy/1.3.0b3/.

geoalchemy2/comparator.py
```diff
@@ -86,7 +89,15 @@ def __getattr__(self, name):
...
-        if not name.startswith('ST_'):
+        if (
+            (
+                SA_VERSION < [1, 3, 4]
+                and not name.startswith('ST_')
+            ) or (
+                SA_VERSION >= [1, 3, 4]
+                and not name.lower().startswith('st_')
+            )
+        ):
```
elemoine: Why not using `if not name.lower().startswith('st_')` with all versions of SQLAlchemy? What problem could that cause?

I tend to think that it's ok to use `if not name.lower().startswith('st_')` with all versions of SQLAlchemy. With SQLAlchemy 1.3.3 and lower it would mean that we accept both ST_ and st_ functions. I think it's ok, even though the st_ (lower-case) version is not registered, meaning that it may not have the correct return type. I think it does no harm to still accept it here.

What do you think? If possible I'd like to avoid SQLAlchemy version conditional code in GeoAlchemy.

elemoine: And for the record, this test is a bit of a hack, and it makes it impossible to use the form Class.geom.spatial_func to call SpatiaLite functions that don't have the ST_ prefix.

geoalchemy2/comparator.py
adrien-berchet: Yes sorry, I was wondering if I should introduce version condition in the code so I chose the dirty way as a first try.

geoalchemy2/comparator.py
adrien-berchet: > What do you think? If possible I'd like to avoid SQLAlchemy version conditional code in GeoAlchemy.

I would also prefer to avoid version condition. I am ok to delegate the case-sensitivity to SQLAlchemy and not fixing it in geoalchemy2. But since we know it is broken for `sqlalchemy < 1.3.4` maybe we should add a comment in the code or in the doc?

> And for the record, this test is a bit of a hack, and it makes it impossible to use the form Class.geom.spatial_func to call SpatiaLite functions that don't have the ST_ prefix.

Indeed. Maybe we should have a kind of `functions._registry` which could be a `set()` containing all the function names? And the test would just become `if name.lower() not in functions._registry`?

geoalchemy2/comparator.py
elemoine: No problem. I questioning the introduction of SQLAlchemy version-dependent code anyway.

geoalchemy2/comparator.py
elemoine: > Indeed. Maybe we should have a kind of `functions._registry` which could be a `set()` containing all the function names? And the test would just become `if name.lower() not in functions._registry`?

Not sure. Because I still want to be able to do Class.geom.ST_Func although I don't have a GenericFunction class for ST_Func. We actually don't need GenericFunction classes for functions that don't return geometries.

elemoine: My suggestion is to use `if not name.lower().startswith('st_')` for all SQLAlchemy versions. What problems do you anticipate with that?

adrien-berchet: > We actually don't need GenericFunction classes for functions that don't return geometries.

Isn't it contradictory to this: https://github.com/geoalchemy/geoalchemy2/pull/224#issuecomment-487983471? ;-P

> Not sure. Because I still want to be able to do Class.geom.ST_Func although I don't have a GenericFunction class for ST_Func.

Ok I see. So let's go for `if not name.lower().startswith('st_')`.
But in this case, what is the point to enumerate the functions that do not return a `Geometry` in `functions.py`?

> My suggestion is to use if not name.lower().startswith('st_') for all SQLAlchemy versions. What problems do you anticipate with that?

I don't anticipate anything more that what we already know, I just feel embarrassed to let a known undesired behavior (returning the wrong type for `sqlalchemy < 1.3.4` when the letter case is not exactly as expected by `geoalchemy2`).

elemoine: > Isn't it contradictory to this: #224 (comment)? ;-P

A little maybe. Currently we add GenericFunction classes for both geometry-returning and non geometry-returning functions. Adding GenericFunction classes for non geometry-returning functions is not that useful, I guess we do it just for documentation purposes.

> But in this case, what is the point to enumerate the functions that do not return a Geometry in functions.py?

Documentation I guess. But this is not a good argument. I think we can revisit that choice.

> I just feel embarrassed to let a known undesired behavior (returning the wrong type for sqlalchemy < 1.3.4 when the letter case is not exactly as expected by geoalchemy2).

But you're not fixing this undesired behavior with this SQLAlchemy version dependent code. Not for every case I mean. There is always the case where the user does func.st_buffer(geom) in her code, as opposed to Class.geom.st_buffer. That case is not detected.

So for SQLAlchemy < 1.3.4 I guess we need to live with the limitation.

elemoine:
@adrien-berchet please merge when you think this is ready.

geoalchemy2/comparator.py
adrien-berchet: > Documentation I guess. But this is not a good argument. I think we can revisit that choice.

Ok. We will see that later then.

> But you're not fixing this undesired behavior with this SQLAlchemy version dependent code. Not for every case I mean. There is always the case where the user does func.st_buffer(geom) in her code, as opposed to Class.geom.st_buffer. That case is not detected.
So for SQLAlchemy < 1.3.4 I guess we need to live with the limitation.

You are right, I can live with that :)

tests/test_functions.py
```diff
@@ -10,21 +13,56 @@
...
+SA_VERSION = parse_version(SA_VERSION)
+
+
 def eq_sql(a, b):
     a = re.sub(r'[\n\t]', '', str(a))
     assert a == b
 
 
 def _test_simple_func(name):
-    eq_sql(getattr(func, name)(1).select(),
```
adrien-berchet: I have a last question. I updated these tests for my dev checks but I am not sure it is relevant to change them as they rely on the private object `sqlalchemy.functions._registry`. Should I revert them?

elemoine: Yep, let's avoid relying on SQLAlchemy private stuff. Thanks.

Issue #220, opened 2019-04-24
How to do a spacial join?


troyswanson:
Consider two objects: points and polygons.

I would like to create a relationship on the point object so that I can see which polygons (if any), the point is contained within.

Writing a query for this in raw SQL is pretty straightforward:

```sql
SELECT *
FROM points
FULL JOIN polygons ON ST_Within(points.the_point, polygons.the_area)
```

How would I represent this relationship in geoalchemy?

elemoine:
Are you using SQLAlchemy Core or ORM? Do you want to configure the join condition in the model or at query time? Also, have you looked at the SQLAlchemy docs? Joining between tables is thoroughly explained in the SQLAlchemy docs...

troyswanson:
Hello @elemoine! Thanks for responding.

I'm using SQLAlchemy ORM, and I'd like to configure the relationship in the model. I've taken a look at the SQLAlchemy docs regarding joins, and from what I can tell, they mostly rely on some kind of foreign key relationship, which this wouldn't really apply to.

The only thing that seems to be relevant is using [custom operators in the join condition](https://docs.sqlalchemy.org/en/13/orm/join_conditions.html#using-custom-operators-in-join-conditions), but I'm having difficulty writing this relationship in code.

elemoine:
It's actually more involved than I expected… And I am suprised I haven't met this before.

I tried this but it does not work:

```py
from sqlalchemy import create_engine, MetaData, Column, Integer
from sqlalchemy.orm import sessionmaker, relationship
from sqlalchemy.ext.declarative import declarative_base
from geoalchemy2 import Geometry
from geoalchemy2 import functions  # NOQA


engine = create_engine("postgresql://elemoine@localhost/gis", echo=True)
metadata = MetaData(engine)

Base = declarative_base(metadata=metadata)


class Polygon(Base):
    __tablename__ = "polygon"
    id = Column(Integer, primary_key=True)
    geom = Column(Geometry("POLYGON", srid=4326))
    points = relationship(
        "Point",
        primaryjoin="func.ST_Contains(foreign(Polygon.geom), remote(Point.geom))",
        viewonly=True,
    )


class Point(Base):
    __tablename__ = "point"
    id = Column(Integer, primary_key=True)
    geom = Column(Geometry("POINT", srid=4326))


metadata.drop_all(checkfirst=True)
metadata.create_all()


polygon = Polygon()
```

I get this error when instantiating `Polygon`:

```
[...]

PR #221, opened 2019-04-25
Support WKBElement as bind values

elemoine:
GeoAlchemy does not currently support WKBElement objects as bind values. We even have a [a comment in the tests](https://github.com/geoalchemy/geoalchemy2/blob/3b4041620907b5cc0a387eb15749d53251e89e77/tests/test_functional.py#L158-L160) for that issue.

Also, Issue #220 has recently revealed problems with lazy spatial relationships because of that lack.

This PR fixes the problem by extending the `bind_processor` function, making it support WKBElement bind-values. The function relies on Shapely for converting the WKBElement into a WKT string. It just throws an exception when Shapely is not available.

### DIFF
--- a/geoalchemy2/types.py
+++ b/geoalchemy2/types.py
@@ -155,10 +155,20 @@ class _GISType(UserDefinedType):
                 else:
                     return 'SRID=%d;%s' % (bindvalue.srid, bindvalue.data)
             elif isinstance(bindvalue, WKBElement):
-                if not SHAPELY:
-                    raise ArgumentError('Shapely required for handling WKBElement bind values')
-                shape = to_shape(bindvalue)
-                return 'SRID=%d;%s' % (bindvalue.srid, shape.wkt)
+                if dialect.name == 'sqlite' or not bindvalue.extended:
+                    # With SpatiaLite or when the WKBElement includes a WKB value rather
+                    # than a EWKB value we use Shapely to convert the WKBElement to an
+                    # EWKT string
+                    if not SHAPELY:
+                        raise ArgumentError('Shapely is required for handling WKBElement bind '
+                                            'values when using SpatiaLite or when the bind value '
+                                            'is a WKB rather than an EWKB')
+                    shape = to_shape(bindvalue)
+                    return 'SRID=%d;%s' % (bindvalue.srid, shape.wkt)
+                else:
+                    # PostGIS ST_GeomFromEWKT works with EWKT strings as well
+                    # as EWKB hex strings
+                    return bindvalue.desc
             else:
                 return bindvalue
         return process
--- a/geoalchemy2/types.py
+++ b/geoalchemy2/types.py
@@ -13,6 +13,13 @@ from sqlalchemy.sql import func
 from sqlalchemy.dialects import postgresql
 from sqlalchemy.dialects.postgresql.base import ischema_names
 
+try:
+    from .shape import to_shape
+    SHAPELY = True
+except ImportError:
+    SHAPELY = False
+
+
 from .comparator import BaseComparator, Comparator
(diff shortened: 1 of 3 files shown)


geoalchemy2/types.py
```diff
@@ -147,6 +154,11 @@ def process(bindvalue):
                     return '%s' % (bindvalue.data)
                 else:
                     return 'SRID=%d;%s' % (bindvalue.srid, bindvalue.data)
+            elif isinstance(bindvalue, WKBElement):
+                if not SHAPELY:
+                    raise ArgumentError('Shapely required for handling WKBElement bind values')
+                shape = to_shape(bindvalue)
+                return 'SRID=%d;%s' % (bindvalue.srid, shape.wkt)
```
adrien-berchet: I think that converting WKB back to WKT is quite disappointing (as far as I observed, WKT can sometimes loose precision and are slower to insert) but I don't know a clean method to use only WKB at this moment. But as soon as `shapely==1.7` is released, I think that this line should be `return wkb.dumps(shape, hex=True, srid=bindvalue.srid)` (see the new `wkb.dumps`: https://github.com/Toblerity/Shapely/blob/master/shapely/wkb.py#L24). Note that this is possible because `ST_GeomFromEWKT()` actually accepts either WKT, EWKT, WKB or EWKB argument value (I only tested with postgis, I don't know how sptialite behaves).

elemoine: Good call. I'll try a bit harder to see if we can avoid the wkt conversion.

I didn't know `ST_GeomFromEWKT` worked with WKB hex strings. This is surprising.

elemoine: Maybe we can merge my code as-is and revisit it when Shapely 1.7 is available. Is that what you're suggesting Adrien ?

adrien-berchet: Yes since I don't know how to do better than this fix without the next release of shapely (except by copying the new code of shapely here, which is event more disappointing) and it is much better than a failure anyway :-)

elemoine: @adrien-berchet I checked, and SpatiaLite's `GeomFromEWKT` does not support (E)WKB hex strings, as expected.

But please have a look at my last commit. There is a case where we can avoid the EWKB to EWKB conversion.

Thank you for letting me know about PostGIS `ST_GeomFromEWKT` being able to work with (E)WKB hex strings.

adrien-berchet: This is nice since the `WKBElement`s returned by queries will be extended, so we should almost always use the `return bindvalue.desc` case when working with PostGIS, except when the `WKBElement`s are explicitely created with `extended=False`, which seems unlikely in real use case.

Note: a test for extended `WKBElement` should be added.

elemoine: > This is nice since the WKBElements returned by queries will be extended, so we should almost always use the return bindvalue.desc case when working with PostGIS, except when the WKBElements are explicitely created with extended=False, which seems unlikely in real use case.

Exactly. That was my thought as well.

> Note: a test for extended WKBElement should be added.

What do you mean? Are you talking about a test related to this PR's additions? Or something else?

adrien-berchet: > What do you mean? Are you talking about a test related to this PR's additions? Or something else?

A test related to this PR. The same as `test_select_bindparam_WKBElement` but with an extended `WKBElement`. Because the `return bindvalue.desc` case seems to be not covered.

elemoine: Ok, will have a look this afternoon.

elemoine: New test added. I'll merge if Travis is green.

elemoine: And I'll create a ticket suggesting the use of `wkb.dumps(srid=xxx)` when Shapely 1.7 is available.

PR #222, opened 2019-04-25
Document the use of spatial functions in primaryjoin conditions

### LINKED ISSUE #220: How to do a spacial join?

elemoine:
Fixes #220.

### DIFF
--- a/doc/orm_tutorial.rst
+++ b/doc/orm_tutorial.rst
@@ -185,14 +185,8 @@ The SQLAlchemy ORM Tutorial's `Querying section
 <http://docs.sqlalchemy.org/en/latest/orm/tutorial.html#querying>`_ provides
 more examples of queries.
 
-Spatial Query
--------------
-
-As spatial database users executing spatial queries is of a great interest to
-us. There comes GeoAlchemy!
-
-Spatial relationship
-~~~~~~~~~~~~~~~~~~~~
+Make Spatial Queries
+--------------------
 
 Using spatial filters in SQL SELECT queries is very common. Such queries are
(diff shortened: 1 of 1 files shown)


elemoine:
I'll merge this this afternoon or tomorrow if I there's no feedback.

troyswanson:
I wonder if it would be helpful to provide an example the other way as well, to show the relationship from the treasure's perspective.

In other words, "what lake does this treasure live in?" vs "what treasure lives in this lake?"

Also, it might be helpful to discuss (very basic) lazy/eager loading strategies. With lazy loading, Iterating on the lake response and accessing the treasure attribute would get very expensive, since it would emit a new SQL query for every lake. Eager loading would eliminate that expense.

Related docs: https://docs.sqlalchemy.org/en/13/orm/loading_relationships.html#joined-eager-loading

elemoine:
I can add a backref to the relationship and a sentence to explain what it does.

I can also mention that the relationship is configured with the default `lazy="select"` strategy, with a link to the appropriate SQLA doc section.

And I think I'll stop there, to avoid rewriting the SQLA docs :)

Thanks for your suggestions!

elemoine:
I added more information. Will merge when the CI process is finished.

PR #223, opened 2019-04-28
Get prepared for 0.6.2

elemoine:
I'd like to release 0.6.2, with the [support of WKBElement objects as bind values](https://github.com/geoalchemy/geoalchemy2/pull/221), which I consider as a bug fix rather than a new feature.


adrien-berchet:
I just added a minor commit to update my account name in the contributors list since it was deprecated.

doc/conf.py
```diff
@@ -49,9 +49,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = '0.6.1'
+version = '0.6.2'
```
sbrunner: The version pattern don't correspond to the comment ...

elemoine: It's ok I think. `version` is the actual the version number, while `release` is the version number + the alpha/beta/rc tag. In our case `version` and `release` are the same, and we use three-digit version numbers.

elemoine:
> I just added a minor commit to update my account name in the contributors list since it was deprecated.

Which I just amended to remove the final dot in the commit message :)

adrien-berchet:
> Which I just amended to remove the final dot in the commit message :)

Ups, sorry and thanks :innocent:

elemoine:
0.6.2 is released on pypi: https://pypi.org/project/GeoAlchemy2/0.6.2/

doc/conf.py
sbrunner: As you want but it doesn't respect the comment :-)

elemoine: Yes, that's a bit error-prone. I'll change it.

sbrunner: :-)

PR #189, opened 2018-08-13
Add suport to some functions

massilva:

PR #224, opened 2019-04-30
Add some missing functions

adrien-berchet:
Add some functions.
(this PR is a revival of #187 , #188 and #189)

### DIFF
--- a/geoalchemy2/functions.py
+++ b/geoalchemy2/functions.py
@@ -100,11 +100,21 @@ _FUNCTIONS = [
     # Geometry Accessors
     #
 
+    ('ST_Boundary', types.Geometry,
+     'Returns the closure of the combinatorial boundary of this Geometry.'),
+
+    ('ST_BoundingDiagonal', types.Geometry,
+     'Returns the diagonal of the supplied geometry\'s bounding box.'),
+
+    ('ST_EndPoint', types.Geometry,
+     'Returns the last point of a ``LINESTRING`` or ``CIRCULARLINESTRING`` '
+     'geometry as a ``POINT``.'),
+
     ('ST_Envelope', types.Geometry,
      'Returns a geometry representing the double precision (float8) bounding'
      'box of the supplied geometry.'),
 
-    ('ST_GeometryN', None,
+    ('ST_GeometryN', types.Geometry,
      'Return the 1-based Nth geometry if the geometry is a '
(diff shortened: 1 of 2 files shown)


geoalchemy2/functions.py
```diff
@@ -139,10 +171,124 @@ def __init__(self, *args, **kwargs):
...
+     'counter-clockwise.'),
+
+    ('ST_ForceRHR', types.Geometry,
+     'Force the orientation of the vertices in a polygon to follow the '
+     'Right-Hand-Rule.'),
+
+    ('ST_ForceSFS', types.Geometry,
+     'Force the geometries to use SFS 1.1 geometry types only.'),
+
+    ('ST_M', types.Geometry,
```
elemoine: Type is not correct.

elemoine:
> While I was writing this PR, I wondered if it would be possible to just ask the DB which functions are returning a `Geometry` and to register these functions automatically, instead of enumerating all of them. Do you think it would be relevant? It seems possible with a query like this one:

Not sure I'd like to introduce something like this. There are a lot of PostGIS functions that don't return a Geometry, so why would we "favor" those that return geometries?

adrien-berchet:
> > While I was writing this PR, I wondered if it would be possible to just ask the DB which functions are returning a `Geometry` and to register these functions automatically, instead of enumerating all of them. Do you think it would be relevant? It seems possible with a query like this one:
> 
> Not sure I'd like to introduce something like this. There are a lot of PostGIS functions that don't return a Geometry, so why would we "favor" those that return geometries?

You are right, this is not the good way of thinking. But maybe querying all the functions related to the PostGIS extension (I don't know the query for this, is it possible?) and getting their return type, or something like this?
I am just wondering but the number of PostGIS functions is not really an issue anyway.

elemoine:
> You are right, this is not the good way of thinking. But maybe querying all the functions related to the PostGIS extension (I don't know the query for this, is it possible?) and getting their return type, or something like this?

It probably is possible, but that would be PostGIS-only.

adrien-berchet:
> It probably is possible, but that would be PostGIS-only.

Ah yes, I always forget that geoalchemy2 also supports spatialite... :)

elemoine:
@adrien-berchet I should have let you merge this. Hope this was complete on your side.

adrien-berchet:
Indeed, you had the finger ready on the trigger for this one ;)
But no problem, it looked ok to me.

PR #226, opened 2019-05-08
Do not register functions.GenericFunction for sqlalchemy>=1.3.4

adrien-berchet:
Do not register `functions.GenericFunction` in `sqlalchemy.sql.functions._registry` for `sqlalchemy>=1.3.4`.
This is not mandatory but it will be more consistent with sqlalchemy.

### DIFF
--- a/geoalchemy2/functions.py
+++ b/geoalchemy2/functions.py
@@ -84,6 +84,8 @@ class GenericFunction(functions.GenericFunction):
             type = Geometry
     """
 
+    _register = False
+
     def __init__(self, *args, **kwargs):
         expr = kwargs.pop('expr', None)
         if expr is not None:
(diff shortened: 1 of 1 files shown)


geoalchemy2/functions.py
```diff
@@ -84,6 +84,8 @@ class ST_TransScale(GenericFunction):
             type = Geometry
     """
 
+    _register = False
```
elemoine: I'd just add a comment explaining the use of this property.

geoalchemy2/functions.py
```diff
@@ -84,6 +84,8 @@ class ST_TransScale(GenericFunction):
             type = Geometry
     """
 
+    # Set _register to False in order to not register this class in
+    # sqlalchemy.sql.functions._registry. Only its childs will be registered.
```
elemoine: s/to not/not to/
s/childs/children/

Thanks

elemoine: I  can do it myself tomorrow.

adrien-berchet: Thanks, I fixed it :)

elemoine:
Please merge.

PR #231, opened 2019-05-16
Do not assume the "public" schema

### LINKED ISSUE #230: Spatial Index assuming public schema
When creating a spatial index for a declaratively defined table that does not specify the schema, `public` is assumed to be the schema for the table (https://github.com/geoalchemy/geoalchemy2/blob/master/geoalchemy2/__init__.py#L81). My issue here is I have the default schema for my DB set up to be `my_schema`, so my table is created in that schema, then when alembic tries to generate the spatial index for the column geoalchemy is prepending a `public.` to the index table, thereby causing the index creation to fail.

elemoine:
This PR aims to fix #230, where GeoAlchemy assumes that the schema is "public" when no schema is specified by the user.

Closes #230.

### DIFF
--- a/geoalchemy2/__init__.py
+++ b/geoalchemy2/__init__.py
@@ -59,7 +59,6 @@ def _setup_ddl_event_listeners():
 
             if event == 'before-drop':
                 # Drop the managed Geometry columns
-                table_schema = table.schema or 'public'
                 for c in gis_cols:
                     if bind.dialect.name == 'sqlite':
                         drop_func = 'DiscardGeometryColumn'
@@ -78,7 +77,6 @@ def _setup_ddl_event_listeners():
             # Restore original column list including managed Geometry columns
             table.columns = table.info.pop('_saved_columns')
 
-            table_schema = table.schema or 'public'
             for c in table.c:
                 # Add the managed Geometry columns with AddGeometryColumn()
                 if isinstance(c.type, Geometry) and c.type.management is True:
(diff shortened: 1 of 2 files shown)


elemoine:
WIP for the moment. Needs testing, and unit-tests probably.

elemoine:
Now with tests…

geoalchemy2/__init__.py
```diff
@@ -105,10 +103,15 @@ def dispatch(event, table, bind):
...
-                                      table.name, c.name))
+                        if table.schema:
+                            bind.execute('CREATE INDEX "idx_%s_%s" ON "%s"."%s" '
+                                         'USING GIST ("%s")' %
+                                         (table.name, c.name, table.schema,
+                                          table.name, c.name))
+                        else:
+                            bind.execute('CREATE INDEX "idx_%s_%s" ON "%s" '
+                                         'USING GIST ("%s")' %
+                                         (table.name, c.name, table.name, c.name))
```
loicgasser: What about a helper function `table_ref(table)`  that returns the string? You could avoid the `if/else` and the duplication of SQL.

elemoine: At the same time the current code is very easy to read and understand. So I tend to think that the duplication is ok here.

elemoine: Loïc, I am going to merge this. But feel free to open a new PR if you think you can improve the code. I'll be happy to merge any improvement. Thanks.

loicgasser: I was just passing by, it was a very superficial look at it.

Issue #198, opened 2018-10-24
Bug in WKT and WKBElement.copy() and serialization problem with eve_sqlalchemy


hectorcanto:
I'm using GeoAlchemy2 0.5.0 in an Eve 0.7.10 (microframework based on Flask) + SQL Alchemy 1.2.11 app.

When serializing my Point Geometry as defined below I get the next error:
`geo = Column(geoalchemy2.Geometry(geometry_type='POINT', srid=4326))`

> TypeError: Object of type Function is not JSON serializable

Debugging the application I've encountered a deep copy of the WKB object in eve_sqlalchemy that returns a function instead of a copy of the object. The copy is to avoid garbage collection of the object, as saids in the eve_sqlalchemy code. The said line is in the following link:

https://github.com/pyeve/eve-sqlalchemy/blob/0.7.0/eve_sqlalchemy/utils.py#L78

The copy method returns a sqlalchemy function (the Function in the TypeError above) instead of a copy of the object. I think there is a problem there in the implementation of WKB (and WKT too).

I believe this might be a bug on the elements. To reproduce it:

```python
from geoalchemy2 import WKTElement
a = WKTElement('POINT(5 45)', srid=4326)
a.copy()
# prompt: <sqlalchemy.sql.functions.Function at 0x7fd3e1693ac8; copy>
```
Anyway, using the debugger I replaced the "function" to a deep copy of the said element and I get another unexpected error which I didn't have the time to explore yet.

> NotImplementedError: Operator 'getitem' is not supported on this expression

So, solving the copy thing might not resolve the situation, so I will appreciate any workaround to tranform the WKB into something serializable.

adrien-berchet:
Thanks for your report.

The first thing is that the `a.copy()` should raise an exception instead of returning a `sqlalchemy.sql.functions.Function` object with no meaning. This behavior is now fixed in the master branch, thanks for having spotted that.

The second thing is about the copy thing. Indeed there is no copy actual method for `WKTElment` or `WKBElement` objects but I am not sure that such method would be relevant, except in your very specific use case. Anyway it is not anticipated in a future release.

[...]

PR #238, opened 2019-06-21
Raise an AttributeError when calling an attribute of _SpatialElement …

adrien-berchet:
Raise an AttributeError when calling an attribute of _SpatialElement that do not starts with the 'st_' prefix, as in the BaseComparator class.

@elemoine do you think it might break something? (it does not break the tests but I am not sure...)

### DIFF
--- a/geoalchemy2/elements.py
+++ b/geoalchemy2/elements.py
@@ -61,6 +61,13 @@ class _SpatialElement(functions.Function):
         # ST_Buffer(ST_GeomFromWKB(:ST_GeomFromWKB_1), :param_1)
         #
 
+        # Function names that don't start with "ST_" are rejected.
+        # This is not to mess up with SQLAlchemy's use of
+        # hasattr/getattr on Column objects.
+
+        if not name.lower().startswith('st_'):
+            raise AttributeError
+
         # We create our own _FunctionGenerator here, and use it in place of
         # SQLAlchemy's "func" object. This is to be able to "bind" the
         # function to the SQL expression. See also GenericFunction above.
--- a/tests/test_functional.py
+++ b/tests/test_functional.py
@@ -256,7 +256,7 @@ class TestSelectBindParam():
         rows = results.fetchall()
         geom = rows[0][1]
         assert isinstance(geom, WKBElement)
-        assert geom.extented
+        assert geom.extended
 
         s = Lake.__table__.select().where(Lake.__table__.c.geom == bindparam('geom'))
         results = self.conn.execute(s, geom=geom)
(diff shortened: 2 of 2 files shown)


elemoine:
@adrien-berchet does this fix an actual bug? Do you have a test-case? Thanks.

adrien-berchet:
#198 spotted that calling `WKTElement("").copy()` returns a `sqlalchemy.sql.functions.Function` object, which has no meaning.

elemoine:
So if I understand correctly this PR would not fix the issue reported with #198, would it?

adrien-berchet:
Not really, it just fixes a undesirable behavior that was revealed in this issue.

By the way, I am wondering if we should add a `copy()` method, as suggested in #198? The reported use case looks very specific, isn't it?

elemoine:
@adrien-berchet As I understand it eve uses `hasattr(val, 'copy')` to determine whether `val` is an association proxy. See https://github.com/pyeve/eve-sqlalchemy/blob/0.7.0/eve_sqlalchemy/utils.py#L77-L78. This sounds rather weak to me.

That being said I agree with the change you're proposing, with a modification to the comment in the code. Indeed, it is not correct to say that "This is not to mess up with SQLAlchemy's use of hasattr/getattr on Column objects" in this case. Instead I'd add this comment : "Raise an AttributeError when the attribute name doesn't start with ST_. This is to be nice with other librairies that use some ducktyping (e.g. hasattr(element, "copy")) to determine the type of the element.".

Also I'd be great to add a unit test for this.

@adrien-berchet tell me if you have time to modify your branch. I can do it if you don't.

Thanks!

adrien-berchet:
> @adrien-berchet As I understand it eve uses `hasattr(val, 'copy')` to determine whether `val` is an association proxy. See https://github.com/pyeve/eve-sqlalchemy/blob/0.7.0/eve_sqlalchemy/utils.py#L77-L78. This sounds rather weak to me.

I agree, that's why I think we can just add this fix and answer in the issue that a `copy()` method is currently not anticipated.

> That being said I agree with the change you're proposing, with a modification to the comment in the code. Indeed, it is not correct to say that "This is not to mess up with SQLAlchemy's use of hasattr/getattr on Column objects" in this case. Instead I'd add this comment : "Raise an AttributeError when the attribute name doesn't start with ST_. This is to be nice with other librairies that use some ducktyping (e.g. hasattr(element, "copy")) to determine the type of the element.".
> 
> Also I'd be great to add a unit test for this.
> 
> @adrien-berchet tell me if you have time to modify your branch. I can do it if you don't.
> 
> Thanks!

Ok. Sorry for this dirty proposal, I wanted your opinion on this before going further.
I won't have any time to clean this PR before next week, so you can of course do it if you have some time earlier (otherwise I think I will do it monday or thursday).

Thanks

elemoine:
@adrien-berchet see https://github.com/geoalchemy/geoalchemy2/pull/240. I'll merge it when Travis is happy.

adrien-berchet:
Perfect, thank you very much!

Issue #213, opened 2019-03-06
WKBElement construct has no default compilation handler


mam8cc:
Using `geoalchemy==0.6.1` appears to introduce a regression for some versions of `SQLAlchemy` (`< 1.1.0`, tested `1.0.12` through `1.0.18`).

```
$ python run.py
Traceback (most recent call last):
  File "/venv/lib/python3.6/site-packages/sqlalchemy/ext/compiler.py", line 456, in __call__
    fn = self.specs['default']
KeyError: 'default'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "run.py", line 56, in <module>
    session.commit()
  File "/venv/lib/python3.6/site-packages/sqlalchemy/orm/scoping.py", line 157, in do
    return getattr(self.registry(), name)(*args, **kwargs)
  File "/venv/lib/python3.6/site-packages/sqlalchemy/orm/session.py", line 801, in commit
    self.transaction.commit()
  File "/venv/lib/python3.6/site-packages/sqlalchemy/orm/session.py", line 392, in commit
    self._prepare_impl()
  File "/venv/lib/python3.6/site-packages/sqlalchemy/orm/session.py", line 372, in _prepare_impl
    self.session.flush()
  File "/venv/lib/python3.6/site-packages/sqlalchemy/orm/session.py", line 2019, in flush
    self._flush(objects)
  File "/venv/lib/python3.6/site-packages/sqlalchemy/orm/session.py", line 2137, in _flush
    transaction.rollback(_capture_exception=True)
  File "/venv/lib/python3.6/site-packages/sqlalchemy/util/langhelpers.py", line 60, in __exit__
    compat.reraise(exc_type, exc_value, exc_tb)
  File "/venv/lib/python3.6/site-packages/sqlalchemy/util/compat.py", line 186, in reraise
    raise value
  File "/venv/lib/python3.6/site-packages/sqlalchemy/orm/session.py", line 2101, in _flush
    flush_context.execute()
  File "/venv/lib/python3.6/site-packages/sqlalchemy/orm/unitofwork.py", line 373, in execute
    rec.execute(self)
  File "/venv/lib/python3.6/site-packages/sqlalchemy/orm/unitofwork.py", line 532, in execute
    uow
  File "/venv/lib/python3.6/site-packages/sqlalchemy/orm/persistence.py", line 174, in save_obj
    mapper, table, insert)
[...]

PR #242, opened 2019-08-20
Add default compilation rules to elements and functions

adrien-berchet:
I tried to understand the origin of #213 and when I add default compilers (as the exception hints) it works properly but breaks many tests for `sqlalchemy < 1.1`, so it may be not the proper way to fix this.

### DIFF
--- a/geoalchemy2/elements.py
+++ b/geoalchemy2/elements.py
@@ -100,8 +100,16 @@ class _SpatialElement(functions.Function):
         raise NotImplementedError()
 
 
+# Default handlers are required for SQLAlchemy < 1.1
+# See more details in https://github.com/geoalchemy/geoalchemy2/issues/213
+@compiles(_SpatialElement)
+def compile_spatialelement_default(element, compiler, **kw):
+    return "{}({})".format(element.name,
+                           compiler.process(element.clauses, **kw))
+
+
 @compiles(_SpatialElement, 'sqlite')
-def compile_spatialelement(element, compiler, **kw):
+def compile_spatialelement_sqlite(element, compiler, **kw):
     return "{}({})".format(element.name.lstrip("ST_"),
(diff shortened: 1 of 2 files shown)


adrien-berchet:
EDIT2: I confirm the tests are failing with sqlalchemy==1.0.19 because the function names are replaced by parameters in the queries (e.g. `:param_1` instead of `:ST_Buffer_2` in `tests/test_types.py:TestGeometry:test_function_call`). But the results of geoalchemy2 queries looks valid in manual tests I made.

adrien-berchet:
@elemoine I have several questions about this PR and would like to have your opinion:
- should we add version conditioning on these changes?
- should we fix the tests for `sqlalchemy < 1.1`?
- should we just add `sqlalchemy >= 1.1` requirement for `geoalchemy2 >= 0.6`?

elemoine:
> should we add version conditioning on these changes?

No, I think it's ok to have a default compilation handler for all the SQLAlchemy versions we support (i.e. SQLAlchemy>=0.8).

> should we fix the tests for sqlalchemy < 1.1

We run the GeoAlchemy tests only against SQLAlchemy 1.1.2 today. And I think it's ok.

I'd just add a comment to the code indicating that a default compilation handler is required with SQLAlchemy < 1.1, with a link to #213 or this PR.

>  should we just add `sqlalchemy >= 1.1` requirement for `geoalchemy2 >= 0.6`?

With your fix GeoAlchemy >= 0.6.4 will work with SQLAlchemy 1.0. So let's keep `SQLAlchemy>=0.8` for now.

elemoine:
@adrien-berchet so let's squash your commits into just one commit and merge your branch into master! And I'll go ahead and release 0.6.4.

Thanks for the work on that.

adrien-berchet:
Thanks for these tips, I updated and merged.

PR #247, opened 2020-01-14
Add __eq__ to wkbelement and make from_shape() able to create extended WKBElement

### LINKED ISSUE #246: WKBElement not comparing as expected
I began by raising this issue over at SQLAlchemy: https://github.com/sqlalchemy/sqlalchemy/issues/5093

(Quoting from that report)
observed behaviour:
i'm using sqlalchemy 1.3.11 with a PostgreSQL/PostGIS backend
i have a table with a Geography column that represents a GIS point
i retrieve row from that table
i update the GIS point _with the same lat/lon values as before_
i observe an unexpected SQL update is issued to the database
(End quote)

Discussion with Mike Bayer @zzzeek suggested that the issue might lie within geoalchemy2, with its root in the comparison of the WKBElement objects.

As it stands, SQLAlchemy is considering my retrieved-from-database geography point to be different from my newly created one, even though they have the same latitude, longitude, srid and WKB

```
attribute geo_point has changes: True
added: 01010000006095875cdb5d184039ecbe63781e4a40
deleted: 01010000006095875cdb5d184039ecbe63781e4a40
```

My belief is that if two WKBElement objects have the same hexlify'd WKB contained within them, then they can be considered to be equal.

Adding methods as follows to WKBElement gives me the behaviour I expect:
```
    def __eq__(self, other):
        return self.desc == other.desc

    def __ne__(self, other):
        return not self.desc == other.desc
```

Is this an acceptable change/fix? Or is there a better approach?

matthew-emw:
fixes #246

### DIFF
--- a/geoalchemy2/elements.py
+++ b/geoalchemy2/elements.py
@@ -58,6 +58,12 @@ class _SpatialElement(HasFunction):
         return "<%s at 0x%x; %s>" % \
             (self.__class__.__name__, id(self), self)  # pragma: no cover
 
+    def __eq__(self, other):
+        return self.desc == other.desc and self.srid == other.srid and self.extended == other.extended
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
     def __getattr__(self, name):
         #
         # This is how things like lake.geom.ST_Buffer(2) creates
@@ -202,12 +208,6 @@ class WKBElement(_SpatialElement):
         if PY3:
             desc = desc.encode(encoding="utf-8")
         return binascii.unhexlify(desc)
-
-    def __eq__(self, other):
-        return self.desc == other.desc
-
-    def __ne__(self, other):
-        return not self.desc == other.desc
 
 
 class RasterElement(_SpatialElement):
--- a/geoalchemy2/elements.py
+++ b/geoalchemy2/elements.py
@@ -203,6 +203,12 @@ class WKBElement(_SpatialElement):
             desc = desc.encode(encoding="utf-8")
         return binascii.unhexlify(desc)
 
+    def __eq__(self, other):
+        return self.desc == other.desc
+
+    def __ne__(self, other):
+        return not self.desc == other.desc
+
 
 class RasterElement(_SpatialElement):
     """
--- a/geoalchemy2/elements.py
+++ b/geoalchemy2/elements.py
@@ -59,7 +59,9 @@ class _SpatialElement(HasFunction):
             (self.__class__.__name__, id(self), self)  # pragma: no cover
 
     def __eq__(self, other):
-        return self.desc == other.desc and self.srid == other.srid and self.extended == other.extended
+        return self.desc == other.desc \
+               and self.srid == other.srid \
+               and self.extended == other.extended
 
     def __ne__(self, other):
         return not self.__eq__(other)
--- a/geoalchemy2/elements.py
+++ b/geoalchemy2/elements.py
@@ -60,8 +60,8 @@ class _SpatialElement(HasFunction):
 
     def __eq__(self, other):
         return self.desc == other.desc \
-               and self.srid == other.srid \
-               and self.extended == other.extended
+            and self.srid == other.srid \
+            and self.extended == other.extended
 
     def __ne__(self, other):
         return not self.__eq__(other)
--- a/geoalchemy2/shape.py
+++ b/geoalchemy2/shape.py
@@ -6,6 +6,7 @@ This module provides utility functions for integrating with Shapely.
     As GeoAlchemy 2 itself has no dependency on `Shapely`, applications using
     functions of this module have to ensure that `Shapely` is available.
 """
+from pkg_resources import parse_version
 
 import shapely.wkb
 import shapely.wkt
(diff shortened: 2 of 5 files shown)


geoalchemy2/elements.py
```diff
@@ -219,6 +219,12 @@ def _data_from_desc(desc):
             desc = desc.encode(encoding="utf-8")
         return binascii.unhexlify(desc)
 
+    def __eq__(self, other):
+        return self.desc == other.desc
```
adrien-berchet: Shouldn't we also compare the `extended` and `srid` values here?
```suggestion
        return self.desc == other.desc and self.srid == other.srid and self.extended == other.extended
```

geoalchemy2/elements.py
```diff
@@ -219,6 +219,12 @@ def _data_from_desc(desc):
...
+
+    def __ne__(self, other):
+        return not self.desc == other.desc
```
adrien-berchet: Maybe just call `__eq__` to avoid writing the test twice?
```suggestion
        return not self.__eq__(other)
```

matthew-emw:
@adrien-berchet Thanks for your help thus far. Unfortunately, this implementation moves us forward, but is incomplete :(

You'll see in the ```test_insert_poi``` test that i've added, i'm making a series of assertions such as
```
assert row[1] == from_shape(Point(1, 1), srid=4326)
```
This is with a view to the comparison between ```WKBElement```s that will be used by SqlAlchemy to determine if the entity has changed.

The Poi class/table uses Geography column type. The Geography column type internally uses ```ST_AsBinary``` when retrieving its representation from the database. As such the comparison we've implemented in ```_SpatialElement.__eq__()``` succeeds, because self.desc is a vanilla WKB representation (which is also used inside ```from_shape()```)

If I try to add a similar comparison to the ```test_insert``` test, it fails. This is because Lake uses Geometry, which internally uses ```ST_AsEWKB```. This means that self.desc is EWKB, and fails the equality test with the WKB from ```from_shape()```

So, I'd appreciate your opinion on next steps: try to do more, potentially needing reimplementation of ```from_shape()``` or the internals of _SpatialElement, or accept this as a partial fix?

adrien-berchet:
Thanks for these changes.
Indeed, the `from_shape()` function would need some update to be able to create extended elements (but it is possible only with recent version of Shapely). I think the best way to solve that would be to make `Geometry` and `Geography` types work the same way. I think they should both use `ST_GeomFromEWKT` and `ST_AsEWKB`, but it would be a breaking change. WDYT @elemoine ?

elemoine:
> WDYT @elemoine ?

I'd like to see where #248 goes first :)

matthew-emw:
@elemoine haha... too right. i'm feeling like i brought a knife to a gun fight :)

adrien-berchet:
Now that #256 is merged, we can come back to this one.

For your issue with the `from_shape()` function, I did the following in a previous project:

``` python
import shapely

if shapely.__version__ < '1.7':
    ######################################################################
    # Duplicate function from Shapely alpha version 1.7a1
    from shapely.geos import WKBWriter, lgeos
    from shapely.geometry.base import geom_factory

    def dumps(ob, hex=False, srid=None, **kw):
        """Dump a WKB representation of a geometry to a byte string, or a
        hex-encoded string if ``hex=True``.

        Parameters
        ----------
        ob : geometry
            The geometry to export to well-known binary (WKB) representation
        hex : bool
            If true, export the WKB as a hexidecimal string. The default is to
            return a binary string/bytes object.
        srid : int
            Spatial reference system ID to include in the output. The default
            value means no SRID is included.
        **kw : kwargs
            See available keyword output settings in ``shapely.geos.WKBWriter``.
        """
        if srid is not None:
            # clone the object and set the SRID before dumping
            geom = lgeos.GEOSGeom_clone(ob._geom)
            lgeos.GEOSSetSRID(geom, srid)
            ob = geom_factory(geom)
            kw["include_srid"] = True
        writer = WKBWriter(lgeos, **kw)
        if hex:
            return writer.write_hex(ob)
        else:
            return writer.write(ob)
    ######################################################################
else:
    from shapely.wkb import dumps  # noqa
```

With this the `from_shape()` function could become something like:
``` python
def from_shape(shape, srid=-1, extended=False):
    if extended:
        return WKBElement(buffer(dumps(shape, srid=srid)), extended=True)
    else:
        return WKBElement(buffer(shape.wkb), srid=srid)
```

adrien-berchet:
I updated the `from_shape()` function so it can return extended WKB elements.
With this you should be able to solve your issue.
Nevertheless, I think we should also make the `Geography` type use extended `WKBElement`s by default. I feel it very weird that `Geometry` use extended WKB (i.e. the WKB also stores the SRID, which is a geographic property) while `Géography` does not... The changes introduced in #256 already imply a major release (at least for the `Raster` type) so I think we can also introduce another breaking change in the `Geography` type. WDYT @elemoine ?

adrien-berchet:
I just realized that exporting the `Geography` as EWKB does not make sense since it is not projected by definition, sorry about that.
So in the end, we could keep this PR as it is.
@matthew-emw could you check that it solve your issue please?

matthew-emw:
@adrien-berchet these changes seem to have solved the issue. SQL UPDATEs are no longer being issued by SQLAlchemy when my geopoints have not changed. many thanks.

adrien-berchet:
Ok, nice. Thanks for these tests.
I will merge this PR soon and I think I will deploy a new release afterwards because the issue you described in #246 is quite annoying.

adrien-berchet:
The last think we have to decide is the default behavior of the `from_shape()` function. On the one hand I think we should keep `extended=False` for backward compatibility considerations, but on the other hand I think that `extended=True` would be more convenient since it is the default behavior of the `Geometry` type...

elemoine:
Thanks for picking that up @adrien-berchet. I'll have a look.

PR #248, opened 2020-01-17
WIP: separate SQL elements from value objects

zzzeek:
The WKTElement, WKBElement, and Raster objects are returned
as value objects from the database by the types
defined in geoalchemy2.types.  similarly, these objects are
established as stateful value objects that are placed into
INSERT, UPDATE etc. statements, and then the _GISType is
responsible for turning these into an appropriate SQL
expression in the bind_expression() method.

However, these value objects are also made to subclass
sqlalchemy.sql.functions.FunctionElement, which means they also
behave as SQL expressions.

This is an architectural mistake as a SQL expression and a data value
are two different things, and SQLAlchemy can't otherwise intelligently
distinguish between an object that is applying both roles leading
to the issue with UPDATE we see at
https://github.com/sqlalchemy/sqlalchemy/issues/5093.

This PR illustrates a partial POC with a very quick way to break out
the "SQL-function-ness" from these value objects, and to instead
store the "SQL function" as an instance variable that is then
consulted by the _GISType.   Basic tests pass except for the one
involving "raster" which seems to be similar but I just wanted to
get the basic idea demonstrated.

Ideally, the WKTElement / WKBElement / Raster objects would behave
as pure value objects and not have any SQL rendering information within
them; the type objects should handle coming up with the correct
functions.Function object for each one at SQL time.

This is also related to #247 which attempted to fix the above
SQLAlchemy issue, except it can't really work because as long as
WK[TB]Element are SQL expressions, the ORM will not treat these
as value objects.

### DIFF
--- a/geoalchemy2/functions.py
+++ b/geoalchemy2/functions.py
@@ -54,7 +54,9 @@ from sqlalchemy.sql import functions
 from sqlalchemy.ext.compiler import compiles
 
 from . import types
+from . import elements
 
+from sqlalchemy import bindparam
 
 class GenericFunction(functions.GenericFunction):
     """
@@ -92,6 +94,14 @@ class GenericFunction(functions.GenericFunction):
         expr = kwargs.pop('expr', None)
         if expr is not None:
             args = (expr,) + args
+
+        args = [
+            bindparam(elem.name, unique=True, value=elem,
+                      type_=types._GISType(from_text=elem.name, name=elem.name))
+            if isinstance(elem, elements.HasFunction) else elem
+            for elem in args
+        ]
+
         functions.GenericFunction.__init__(self, *args, **kwargs)
 
 
--- a/geoalchemy2/elements.py
+++ b/geoalchemy2/elements.py
@@ -14,7 +14,10 @@ from .compat import PY3, str as str_
 from .exc import ArgumentError
 
 
-class _SpatialElement(functions.Function):
+class HasFunction(object):
+    pass
+
+class _SpatialElement(HasFunction):
     """
     The base class for :class:`geoalchemy2.elements.WKTElement` and
     :class:`geoalchemy2.elements.WKBElement`.
@@ -44,7 +47,7 @@ class _SpatialElement(functions.Function):
             args = [self.geom_from_extended_version, self.data]
         else:
             args = [self.geom_from, self.data, self.srid]
-        functions.Function.__init__(self, *args)
+        self.function_expr = functions.Function(*args)
 
     def __str__(self):
         return self.desc
@@ -73,45 +76,36 @@ class _SpatialElement(functions.Function):
         # SQLAlchemy's "func" object. This is to be able to "bind" the
         # function to the SQL expression. See also GenericFunction above.
 
-        func_ = functions._FunctionGenerator(expr=self)
+        func_ = functions._FunctionGenerator(expr=self.function_expr)
         return getattr(func_, name)
+
+    @property
+    def name(self):
+        return self.function_expr.name
 
     def __getstate__(self):
         state = {
             'srid': self.srid,
             'data': str(self),
             'extended': self.extended,
-            'name': self.name,
+            'name': self.function_expr.name,
         }
         return state
 
     def __setstate__(self, state):
-        self.__dict__.update(state)
+        self.srid = state['srid']
+        self.extended = state['extended']
         self.data = self._data_from_desc(state['data'])
-        args = [self.name, self.data]
+        args = [state['name'], self.data]
         if not self.extended:
             args.append(self.srid)
-        # we need to call Function.__init__ to properly initialize SQLAlchemy's
-        # internal states
-        functions.Function.__init__(self, *args)
+        self.function_expr = functions.Function(*args)
 
     @staticmethod
     def _data_from_desc(desc):
         raise NotImplementedError()
 
 
-# Default handlers are required for SQLAlchemy < 1.1
-# See more details in https://github.com/geoalchemy/geoalchemy2/issues/213
-@compiles(_SpatialElement)
-def compile_spatialelement_default(element, compiler, **kw):
-    return "{}({})".format(element.name,
-                           compiler.process(element.clauses, **kw))
-
-
-@compiles(_SpatialElement, 'sqlite')
-def compile_spatialelement_sqlite(element, compiler, **kw):
-    return "{}({})".format(element.name.lstrip("ST_"),
-                           compiler.process(element.clauses, **kw))
 
 
 class WKTElement(_SpatialElement):
@@ -220,7 +214,7 @@ class WKBElement(_SpatialElement):
         return binascii.unhexlify(desc)
 
 
-class RasterElement(FunctionElement):
+class RasterElement(HasFunction):
     """
     Instances of this class wrap a ``raster`` value. Raster values read
     from the database are converted to instances of this type. In
@@ -232,7 +226,7 @@ class RasterElement(FunctionElement):
 
     def __init__(self, data):
         self.data = data
-        FunctionElement.__init__(self, self.data)
+        self.function_expr = functions.Function(self.name, self.data)
 
     def __str__(self):
         return self.desc  # pragma: no cover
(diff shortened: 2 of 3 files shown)


elemoine:
Thanks @zzzeek. I hope to look at your work (and possibly pick it up) at the end of next week. I won't be able to do it before.

zzzeek:
not sure if you'd remember but I think I gave you a proof of concept years ago on this, and...I'm very sorry if this was all my fault in the first place :)

elemoine:
No problem! I don't remember anyway :)

geoalchemy2/functions.py
```diff
@@ -92,6 +94,14 @@ def __init__(self, *args, **kwargs):
         expr = kwargs.pop('expr', None)
         if expr is not None:
             args = (expr,) + args
+
+        args = [
+            bindparam(elem.name, unique=True, value=elem,
+                      type_=types._GISType(from_text=elem.name, name=elem.name))
+            if isinstance(elem, elements.HasFunction) else elem
+            for elem in args
+        ]
```
elemoine: I wonder why this is not simply:

```py
    args = [elem.function_expr if isintance(elem, elements.HasFunction) else elem for elem in args]
```

?

elemoine:
I took a brief look at this tonight. This is not easy. I need to wrap my head around it.

elemoine:
In particular, I don't know how to handle the "raster" case for now. I am not the author of the "raster" part, which makes it harder. Anyway, I hope to get back to this soon.

elemoine:
For `RasterElement` I tried to add a `cast_expr` (in the same way @zzzeek added a `function_expr` to `_SpatialElement`), but `cast` takes a type as the second argument, which would lead to a circular dependency between the `elements` and `types` modules.

zzzeek:
I think let's try to make sure the more traditional geoaclhemy constructs do what is needed here then we'd have to try to fit "Raster" into that also.

geoalchemy2/functions.py
zzzeek: I think it could be that also.  but the way we have SQL concepts sprinkled around in multiple places is part of what makes this hard to follow.

elemoine:
> I think let's try to make sure the more traditional geoaclhemy constructs do what is needed here then we'd have to try to fit "Raster" into that also.

I just pushed the `ele_elements` branch to the repo. This branch is based on your change, with simplifications I made while going over your changes and understanding them, and with the raster-related code completely removed to ease the testing of the "traditional" geoalchemy constructs. The tests pass, except for the Spatialite tests that are broken by the removal of the `compile_spatialelement_sqlite` function.

zzzeek:
I notice all the tox.ini stuff is hardcoded to SQLAlchemy 1.1 and this is part of what's making it difficult for me to run all tests, and if I just run "py.test", it doesn't run the test_functional_spatialite stuff anyway I guess that's because I have to install new things.

in tox.ini for alembic I use tox [generative environments](https://tox.readthedocs.io/en/latest/config.html#generating-environments-conditional-settings) which would make this more open ended, in https://github.com/sqlalchemy/alembic/blob/master/tox.ini you can see under "deps" I have listed factors sqla11, sqla12, sqla13, sqlamaster; that dynamically alters the SQLAlchemy dependency based on a token so I can run:

tox -e py37-sqla12
tox -e py37-sqla13
tox -e py37-sqlamaster

etc. 

then in my jenkins CI I do a matrix job that has these all in separate sub-builds.

zzzeek:
now as far as the `@compiles` for the base elements, if you can show me the stack trace for that failure I might be able to spot it, but generally, we're looking for there to be a separation between **value** objects and **sql expression** objects.  That is, if we say the WKBElement is a value object, which is what we're doing at the moment, then I would never want to "compile" this value object.

the reason WKBElement seems to be a "value" object is because, at the moment, the bind_processor and result_processor of the _GISType are receiving and returning WKBElement objects as though they were values.

We could just as well leave everything here in place as it was, and instead create new value objects that look like a WKBElement / SpatialElement except they aren't SQL functions.    

it's just like if I have a SQLite database and I deal with the DateTime column type, it receives and returns Python datetime() objects.  But I can't "compile" a Python datetime() object, whenever a datetime object gets into a SQL statement there's a "literal coercion" step that takes place.   Like if I said this:

   some_table.c.some_date == datetime.datetime.today()

the `BinaryExpression` you get back will have a `bindparam()` object on the right side that is containing this new datetime object.  that is, it gets wrapped in a SQL construct that can accommodate it.   that's what we'd want to go for here as well.

Also I don't want to make the impression that this is all "urgent", as we've seen, the only real side effect of this is that the ORM is not able to be intelligent about a SpatialElement object because it's unclear if it should be treated as a value or a SQL expression.    There is the possibility that the ORM could have new hooks that geoalchemy could use to tell it which for a specific case (basically if the SpatialElement is stateful or not), but this means SQLAlchemy has to add support for this hook everywhere to work around what really should be clear from the external side.

elemoine:
> now as far as the `@compiles` for the base elements, if you can show me the stack trace for that failure I might be able to spot it, but generally, we're looking for there to be a separation between **value** objects and **sql expression** objects. That is, if we say the WKBElement is a value object, which is what we're doing at the moment, then I would never want to "compile" this value object.

Yes. Primarily WKBElement objects are value objects.

We made them sql expressions as well to be able to do things like `wkbelement.ST_Buffer(2)` and `func.ST_Buffer(wkbelement, 2)`, and have SQLA compile this to `ST_Buffer(ST_GeomFromWKB(value), 2)`.

Plus, having them be sql expressions made it possible to compile them differently in the case of SQLite/SpatiaLite. (The `compile_spatialelement_sqlite` you had to remove in your patch.)

> We could just as well leave everything here in place as it was, and instead create new value objects that look like a WKBElement / SpatialElement except they aren't SQL functions.
> 
> it's just like if I have a SQLite database and I deal with the DateTime column type, it receives and returns Python datetime() objects. But I can't "compile" a Python datetime() object, whenever a datetime object gets into a SQL statement there's a "literal coercion" step that takes place. Like if I said this:
> 
> some_table.c.some_date == datetime.datetime.today()
> 
> the `BinaryExpression` you get back will have a `bindparam()` object on the right side that is containing this new datetime object. that is, it gets wrapped in a SQL construct that can accommodate it. that's what we'd want to go for here as well.

Yes, but how? What do we need to do for this "literal coercion" to actually happen? I think that's the piece of the puzzle that I'm missing here.

> Also I don't want to make the impression that this is all "urgent", as we've seen, the only real side effect of this is that the ORM is not able to be intelligent about a SpatialElement object because it's unclear if it should be treated as a value or a SQL expression.

Yes, I understand. I'd like to see this fixed in GeoAlchemy. Thanks for the time you spend on GeoAlchemy! I wish I could do the same on SQLAlchemy.

> There is the possibility that the ORM could have new hooks that geoalchemy could use to tell it which for a specific case (basically if the SpatialElement is stateful or not), but this means SQLAlchemy has to add support for this hook everywhere to work around what really should be clear from the external side.

Let's see if we can change GeoAlchemy first. I just don't see how for the moment, but I am confident we/you'll find a way :)

elemoine:
> and if I just run "py.test", it doesn't run the test_functional_spatialite stuff anyway I guess that's because I have to install new things.

Yes, this is all documented in [`TESTS.rst`](https://github.com/geoalchemy/geoalchemy2/blob/master/TEST.rst).

Regarding you tox-related comments I'll create a separate issue for that.

elemoine:
> Yes, but how? What do we need to do for this "literal coercion" to actually happen? I think that's the piece of the puzzle that I'm missing here.

Use `type_coerce()` in `bind_expression` maybe?

zzzeek:
the usual way that a particular Python literal gets coerced is that it is in some way associated with the TypeEngine that handles it.

the ways in which this occur include:

1. you put the value in an insert() or update() construct associated with a Column that has that type.   or use the ORM that ultimately does the same thing.

2. if you put it in the WHERE clause of a SELECT, or otherwise generated some kind of SQL expression that uses operators (i.e. ==, !=, +, etc.)  the Core expression system coerces it at the point at which you compare the object to a column or expression of that type (again the ORM uses this same system as well):

    widget = some_plain_value()
    some_column = Column('foo', MyType())
    expression = some_column == widget    # <--- MyType() is now associated with "widget"

Some background on the "coercion" system in terms of TypeDecorator, but applies to all the types, is at https://docs.sqlalchemy.org/en/13/core/custom_types.html#types-custom .
3. as you noted, using explicit approaches like type_coerce(value, type), bindparam(value, type_=type), literal(value, type_=type) etc.   but in practice this is not usually needed.

4. just to reduce confusion, SQLAlchemy has some "hardcoded" type coercions built in for simple types like string, int, float, boolean, datetimes.   these have the effect that if you say bindparam(12), it automatically figures out that the type is Integer.    However these are "nice to haves" and you can get by without these, I bring this up just so you aren't confused that these types of data values seem to be coerced more eagerly.

elemoine:
Thanks for that @zzzeek.

adrien-berchet:
@elemoine @zzzeek
I tried to continue what you started here in #256. I achieved a working version with the following properties:
* all elements (sub-classes of _SpatialElement) are purged from any SQL concept, except the `__getattr__` method to allow calls like `elem.ST_Buffer(2)`.
* Raster type and RasterElement class are now just sub-classes of _GISType and _SpatialElement (though I had to slightly update their behavior).

I think it is close to what you had in mind but I would like to know your opinion about it.

elemoine:
Thanks a lot @adrien-berchet. I'll look at it today!

adrien-berchet:
Thanks a lot for this @zzzeek and @elemoine , it was fixed by #256

PR #253, opened 2020-03-10
Update test matrix removing Python 3.4 and adding 3.6, 3.7, 3.8

cjmayo:
Python 3.4 is now EOL.

---

I've got the tests working on 3.{6,7,8}  locally, with 0.6.3,  plus e05a676350b11f0e73 because I have PostGIS 3.0.1 and SQLAlchemy 1.2.17 because of PostgreSQL 12. Let's see how it goes here, maybe a newer SQLAlchemy will be needed for newer Pythons.


cjmayo:
Python 3.7 and 3.8 not available on Trusty. Pypy also failing.

cjmayo:
All versions of Python now working, Reduced the overall build time to something similar to the current by using libsqlite3-mod-spatialite package instead of compiling from source.

PyPy still failing.

adrien-berchet:
Thanks for this work, the CI was indeed quite outdated.
If you have some time to fix PyPy it will be perfect.

cjmayo:
I don't have any ideas at the moment why PyPy is failing but will keep looking.

Broken again because flake8 was not Python 3.8 compatible (pyflakes specifically).

cjmayo:
OK. That's working for Python and PyPy now - but it is using a snapshot of Shapely to get PyPy working.
I guess that means waiting for Shapely 1.7.1.

I did also add an update to the classifier in setup.py. Happy to take that off.

requirements.txt
```diff
@@ -2,9 +2,9 @@
...
 # Additional requirements for running the testsuite and development
-pycodestyle==2.2.0
-flake8==3.2.0
+flake8==3.7.9
 pytest==3.7.4
 pytest-cov==2.5.1
 
-Shapely>=1.3.0
+#Shapely>=1.3.0
+https://github.com/Toblerity/Shapely/archive/9ca22240d3272df9598c30878c6f8562e50f1ec8.tar.gz
```
adrien-berchet:
```suggestion
Shapely>=1.3.0, <1.7
```
I think we could just not use Shapely 1.7 in the CI for now (Shapely is not a mandatory dependency of `geoalchemy2`). Anyway, the `to_shape` and `from_shape` functions only use very basic IO features of Shapely which are very stable.

adrien-berchet:

PR #256, opened 2020-03-26
Continuation of Pr248

adrien-berchet:
Following #248, I made some tests to try to make it work.
It seems ok except for PyPy.

### DIFF
--- a/geoalchemy2/elements.py
+++ b/geoalchemy2/elements.py
@@ -211,6 +211,7 @@ class RasterElement(_SpatialElement):
     most cases you won't need to create ``RasterElement`` instances
     yourself.
     """
+
     geom_from_extended_version = 'raster'
 
     def __init__(self, data):
--- a/geoalchemy2/elements.py
+++ b/geoalchemy2/elements.py
@@ -17,6 +17,7 @@ from .exc import ArgumentError
 class HasFunction(object):
     pass
 
+
 class _SpatialElement(HasFunction):
     """
     The base class for :class:`geoalchemy2.elements.WKTElement` and
@@ -76,7 +77,7 @@ class _SpatialElement(HasFunction):
         # SQLAlchemy's "func" object. This is to be able to "bind" the
         # function to the SQL expression. See also GenericFunction above.
 
-        func_ = functions._FunctionGenerator(expr=self.function_expr)
+        func_ = functions._FunctionGenerator(expr=self)
         return getattr(func_, name)
 
     @property
@@ -104,8 +105,6 @@ class _SpatialElement(HasFunction):
     @staticmethod
     def _data_from_desc(desc):
         raise NotImplementedError()
-
-
 
 
 class WKTElement(_SpatialElement):
--- a/geoalchemy2/elements.py
+++ b/geoalchemy2/elements.py
@@ -50,14 +50,6 @@ class _SpatialElement(HasFunction):
         self.srid = srid
         self.data = data
         self.extended = extended
-        if self.extended:
-            func_name = self.geom_from_extended_version
-            args = [self.data]
-        else:
-            func_name = self.geom_from
-            args = [self.data, self.srid]
-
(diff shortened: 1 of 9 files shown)


tests/test_functional.py
```diff
@@ -354,12 +316,14 @@ def setup(self):
...
     def teardown(self):
         session.rollback()
+        session.expunge_all()
         metadata.drop_all()
 
     def _create_one_lake(self):
         lake = Lake(WKTElement('LINESTRING(0 0,1 1)', srid=4326))
         session.add(lake)
         session.flush()
+        session.expunge_all()
```
adrien-berchet: @elemoine , @zzzeek I had to add several `session.expunge_all()` calls to make the tests pass in PyPy... but I don't understand why it is required now while it was not before. Do you have any clue?

adrien-berchet:
I tried to refactor the Raster and RasterElement classes. The Raster is now very close to the Geometry type and elements.

geoalchemy2/functions.py
```diff
@@ -92,7 +93,20 @@ def __init__(self, *args, **kwargs):
...
+                if elem.extended:
+                    func_name = elem.geom_from_extended_version
+                    func_args = [elem.data]
+                else:
+                    func_name = elem.geom_from
+                    func_args = [elem.data, elem.srid]
+                new_arg = getattr(functions.func, func_name)(*func_args)
+            else:
+                new_arg = elem
+            new_args.append(new_arg)
```
elemoine, line 108: Or maybe a bit more elegantly:
```py
args = list(args)
if expr is not None:
    args = [expr] + args
for idx, elem in enumerate(args):
    if isinstance(elem, elements.HasFunction):
        if elem.extended:
            func_name = elem.geom_from_extented_version
            func_args = [elem.data]
        else:
            func_name = elem.geom_from
            func_args = [elem.data, elem.srid]
        args[idx] = getattr(functions.func, func_name)(*func_args)
functions.GenericFunction.__init__(self, *args, **kwargs)
```

tests/test_functional.py
elemoine: I have no clue unfortunately. I see nothing in your changes that could have an impact on that.

elemoine:
Thanks a lot @adrien-berchet. That's great work! I have no idea about the need for `expunge_all` calls with PyPy. That's rather strange.

geoalchemy2/elements.py
```diff
@@ -81,39 +85,19 @@ def __getstate__(self):
             'srid': self.srid,
             'data': str(self),
             'extended': self.extended,
-            'name': self.name,
         }
         return state
 
     def __setstate__(self, state):
-        self.__dict__.update(state)
```
elemoine, line 89: That might be the cause of the behavior change with PyPy. I am not fully sure, but that's my best intuition for the moment.

tests/test_functional.py
elemoine: @adrien-berchet I've just added a commit that removes the call to `expunge_all`, in favor to calls to `expire`, which, I think makes more sense, and `expire` is what's used elsewhere in this test file.

geoalchemy2/functions.py
elemoine: I've added a commit for that.

elemoine:
I think this is good to go @adrien-berchet. As for PyPy I think the calls to `expire` in `test_functional.py` make sense, although I am not sure why the behavior is different with CPython and PyPy.

geoalchemy2/functions.py
adrien-berchet: It is much better indeed, thanks :-)

adrien-berchet:
Thanks @elemoine !
I added some comments so I will merge when the CI is complete.

geoalchemy2/functions.py
elemoine: I'd say "a bit nicer" :-)

Issue #251, opened 2020-02-16
disable auto wrapped st_AsEWKB


DeoLeung:
Hi, I found that using the library will wrapped `st_asewkb` automatically around some postgis function and geometry column, is there a way to disable it?

this causes some type mis-match since we need to compile it into raw sql and pass to other application.

loicgasser:
Wrapping most likely happens here:
https://github.com/geoalchemy/geoalchemy2/blob/master/geoalchemy2/types.py#L148

Depending on what you want to do you can change the wrapping function by sub-classing `Geometry` (at your own risk):
https://github.com/geoalchemy/geoalchemy2/blob/master/geoalchemy2/types.py#L219
You just need to change the `as_binary` class attribute to whatever you need.

adrien-berchet:
Hi
As @loicgasser said, if you want to change to behavior of the geometry you can define a custom geometry class with the preferred function to use. For example, if you want to use `ST_AsGeoJson` instead of the default `ST_AsEWKB` you can define the following class to override the `as_binary` static attribute:

```
class GeoJsonGeometry(Geometry):
    as_binary = 'ST_AsGeoJson'
```

But your use case seems very specific if you don't want to execute the query but just building it with geoalchemy2 and then executing it with another application. You can not just set the `as_binary` attribute to None or an empty string to achieve this, you have to override the `column_expression` method of your custom class. Here is an example that shows the two methods:

```
from geoalchemy2 import Geometry
from sqlalchemy import Table, Column, MetaData, Integer, create_engine, select, insert


class GeoJsonGeometry(Geometry):
    as_binary = 'ST_AsGeoJson'


class RawGeometry(Geometry):

    def column_expression(self, col):
        return col


engine = create_engine('postgresql://adrien@localhost/test_251')

conn = engine.connect()
metadata = MetaData()

conn.execute("CREATE EXTENSION IF NOT EXISTS postgis;")

the_table = Table(
    'the_table', metadata,
    Column('id', Integer, primary_key=True),
    Column('bin_geom', Geometry('LINESTRING', 4326)),
    Column('geojson_geom', GeoJsonGeometry('LINESTRING', 4326)),
    Column('raw_geom', RawGeometry('LINESTRING', 4326)),
)

the_table.drop(conn, checkfirst=True)
the_table.create(conn)

data = [
    {
        "id": 0,
        "bin_geom": "SRID=4326;LINESTRING (0 0, 1 0)",
[...]

Issue #244, opened 2019-08-23
Is it possible to use commands like St_Length on insert?


hyperknot:
I am inserting rows in a database with a column for GeoJSON and calculating the length in a separate update step. The update code is like the following:

```
update table set distance = St_Length(ST_GeomFromGeoJSON(route->>'geometry')::geography, false) where distance is NULL where ...
```

Since the update code takes quite a while to run (can be up to 10 minutes), I'd like to make this calculation done as part of the insert, not as as a separate update. 

Is this possible with Geoalchemy2? I am using SQLAlchemy Core + PostgreSQL 9.6 + PostGIS 2.3.0.

elemoine:
@hyperknot, does the following answer your question?

```py
from sqlalchemy import create_engine, MetaData, Table, Column, Integer, Float, func
from geoalchemy2 import Geometry
from geoalchemy2 import functions  # NOQA


metadata = MetaData()

t = Table(
    "inserts",
    metadata,
    Column("id", Integer, primary_key=True),
    Column("geom", Geometry("LINESTRING", 4326)),
    Column("distance", Float),
)


engine = create_engine("postgresql://elemoine@localhost/gis", echo=True)

t.drop(engine, checkfirst=True)
t.create(engine)

values = [("SRID=4326;LINESTRING(0 0,1 0)", 1.0), ("SRID=4326;LINESTRING(0 0,0 1)", None)]

for value in values:
    if value[1] is None:
        d = func.ST_Length(func.ST_GeomFromEWKT(value[0]))
    else:
        d = value[1]
    i = t.insert().values(geom=value[0], distance=d)
    engine.execute(i)
```

Feel free if that makes no sense and that I haven't understood your question.

hyperknot:
Thanks, this helped a lot. My problem is that it doesn't work in batch insert mode.

```
from sqlalchemy import MetaData, Table, Column, Integer, Float, func
metadata = MetaData()

t = Table("inserts", metadata, Column("id", Integer, primary_key=True), Column("distance", Float))

t.drop(pg_engine, checkfirst=True)
t.create(pg_engine)

# works
d = func.ST_Length(func.ST_GeographyFromText("SRID=4326;LINESTRING(0 0,1 0)"))
i = t.insert().values(distance=d)
pg_engine.execute(i)

# doesn't work
values = [{'distance': func.ST_Length(func.ST_GeographyFromText("SRID=4326;LINESTRING(0 0,1 0)"))}]

[...]

Issue #219, opened 2019-04-03
transforming srid on insert/update


stev-0:
Is there any function that will correctly write the necessary SQL? I assume you could do this with the SQL Statements API and use ST_Transform, but I would much rather use the ORM - is this available at all? This is often required when converting from web maps to company-stored data.

adrien-berchet:
To my knowledge there is not such a function indeed. But in my opinion, such a function would not be consistent with the philosophy of `geoalchemy2` and `sqlalchemy` which encourage to write explicit queries and to reduce the 'magic' as much as possible.

elemoine:
If I had to do that I think I'd use a `TypeDecorator`. Something like this:

```py
from sqlalchemy import create_engine, MetaData, Column, Integer, func, TypeDecorator
from sqlalchemy.orm import sessionmaker
from sqlalchemy.ext.declarative import declarative_base
from geoalchemy2 import Geometry
from geoalchemy2 import functions  # NOQA


engine = create_engine("postgresql://elemoine@localhost/gis", echo=True)
metadata = MetaData(engine)

Base = declarative_base(metadata=metadata)


class TransformedGeometry(TypeDecorator):
    impl = Geometry

    def __init__(self, db_srid, app_srid, **kwargs):
        kwargs["srid"] = db_srid
        self.impl = self.__class__.impl(**kwargs)
        self.app_srid = app_srid

    def column_expression(self, col):
        return self.impl.column_expression(func.ST_Transform(col, self.app_srid))

    def bind_expression(self, bindvalue):
        return func.ST_Transform(self.impl.bind_expression(bindvalue), self.impl.srid)


class Point(Base):
    __tablename__ = "point"
    id = Column(Integer, primary_key=True)
    geom = Column(TransformedGeometry(db_srid=2154, app_srid=4326, geometry_type="POINT"))


metadata.drop_all(checkfirst=True)
metadata.create_all()

p = Point()
p.geom = "SRID=4326;POINT(5 45)"

Session = sessionmaker(bind=engine)
session = Session()

session.add(p)
session.commit()

session = Session()
for point in session.query(Point):
    print(session.scalar(func.ST_AsEWKT(point.geom)))
```

@stev-0 curious to know if that would work for you.

PR #260, opened 2020-04-29
Add a gallery with tests and doc based on examples given in #219, #244 and #251

adrien-berchet:
We had several questions about it so I think it can be good to add some doc on how to use TypeDecorator

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -75,6 +75,18 @@ system. If you're new to GeoAlchemy 2 start with this.
    core_tutorial
    spatialite_tutorial
 
+
+Gallery
+---------
+
+.. toctree::
+   :hidden:
+
+   gallery/index
+
+The :ref:`gallery` page shows examples of the functionalities of GeoAlchemy 2.
+
+
 Reference Documentation
 -----------------------
 
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -84,7 +84,7 @@ Gallery
 
    gallery/index
 
-The :ref:`gallery` page shows examples of the functionalities of GeoAlchemy 2.
+The :ref:`gallery` page shows examples of the GeoAlchemy 2's functionalities.
 
 
 Reference Documentation
--- a/doc/orm_tutorial.rst
+++ b/doc/orm_tutorial.rst
@@ -361,19 +361,6 @@ measurement functions can be applied to
     >>> print '%s: %f' % (lake.name, bufferarea)
     Majeur: 21.485781
 
-Automatically use a function at insert or select
-------------------------------------------------
-
-Sometimes the application always want to apply a function in an insert or in a select.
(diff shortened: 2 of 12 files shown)


adrien-berchet:
@elemoine I wonder if we could not make a gallery in the doc to collect examples of this kind, so when we provide a new example in an issue we can add it to the gallery. I think it would be easier for people to find relevant examples if they can find them in the doc instead of having digging in old issues.

elemoine:
> @elemoine I wonder if we could not make a gallery in the doc to collect examples of this kind, so when we provide a new example in an issue we can add it to the gallery. I think it would be easier for people to find relevant examples if they can find them in the doc instead of having digging in old issues.

Yep, I fully support that idea! Sounds great.

doc/orm_tutorial.rst
```diff
@@ -361,6 +361,19 @@ measurement functions can be applied to
     >>> print '%s: %f' % (lake.name, bufferarea)
     Majeur: 21.485781
 
+Automatically use a function at insert or select
+------------------------------------------------
+
+Sometimes the application always want to apply a function in an insert or in a select.
```
elemoine, line 367: Having both "sometimes" and "always" here sounds strange to.

Also s/want/wants

adrien-berchet:
Thanks, I fixed according to your comments.
I also setup the gallery and moved the example to it.

adrien-berchet:
@elemoine I added two more examples so I re-request a review before merging this.
Also, how do you publish the new doc on readthedoc?

doc/index.rst
```diff
@@ -75,6 +75,18 @@ system. If you're new to GeoAlchemy 2 start with this.
...
+
+Gallery
+---------
+
+.. toctree::
+   :hidden:
+
+   gallery/index
+
+The :ref:`gallery` page shows examples of the functionalities of GeoAlchemy 2.
```
elemoine, line 87: I'd say "of GeoAlchemy 2's functionality".

tests/gallery/test_disable_wrapping.py
```diff
@@ -0,0 +1,64 @@
+"""
+Disable wrapping in select
+==========================
+
+If the application wants to build queries with GeoAlchemy 2 and get them as strings,
```
elemoine, line 5: s/get/gets

tests/gallery/test_disable_wrapping.py
```diff
@@ -0,0 +1,64 @@
...
+the wrapping of geometry columns by a `ST_AsEWKB()` function might be annoying. In
```
elemoine, line 6: s/by/with

elemoine:
A few more comments, but it looks great. Feel free to merge whenever you think it's ready. Thanks a lot.

adrien-berchet:
Ok, I will try to merge all the current PR today so we can release a new version.
Thanks!

PR #262, opened 2020-05-13
Fix requirements

adrien-berchet:
We forced `shapely<1.7` in requirements because it fails with Pypy but now it is only restricted for Pypy.


elemoine:
Looks great.

PR #296, opened 2020-10-30
Update setup.py to point to new home page location

EdwardBetts:


adrien-berchet:
Indeed, thank you very much!

PR #298, opened 2020-11-05
Add many missing functions

adrien-berchet:
I am wondering: some functions return types we do not support (box types and composite types). Should we add them? Or should we don't add these functions and let users defines them according to their needs?
WDYT @elemoine ?


jaddison:
I'm curious what functions were added here?

It's hard to tell as it looks like The code was moved around.

(I'm here to see if ST_distancespheroid existed in the project before this pull request existed - it looks like it was added in this PR? Or maybe ST_distance is actually what I wanted?)

jaddison:
Just following up on my previous comment. Based on information from here: http://postgis.net/workshops/postgis-intro/geography.html#function-list

> `ST_Distance(geometry, geometry)`: For geometry type Returns the 2-dimensional Cartesian minimum distance (based on spatial ref) between two geometries in projected units. **For geography type defaults to return spheroidal minimum distance between two geographies in meters**

Sounds like this is the function I need, and it's already available in geoalchemy without this PR?

adrien-berchet:
Hi @jaddison 
Before this PR, the functions `ST_Distance` and `ST_Distance_Sphere` were available. This PR adds the `ST_DistanceSpheroid` function, among many others.
If you need it now, you can define it yourself as explained here: https://geoalchemy-2.readthedocs.io/en/latest/spatial_functions.html#geoalchemy2.functions.GenericFunction

Issue #310, opened 2021-02-09
Functions returning NULL instead of element data


jwtrhs:
I believe this is the source of the problem in #151

We have the following failing test in our code:

```python
wkb = from_shape(box(1, 1, 2, 2), srid=4326)
query = session.query(GeoTable).filter(func.ST_INTERSECTS(Catchment.geometry, wkb))

assert query.sql == (
    "SELECT geotable.id \n"
    "FROM geotable \n"
    "WHERE ST_Intersects(geotable.geometry, '0103000000010000000500000000000000000"
    "00040000000000000f03f00000000000000400000000000000040000000000000f03f000000000"
    "0000040000000000000f03f000000000000f03f0000000000000040000000000000f03f')"
)
```

which fails on the WHERE clause because `query.sql` returns `WHERE ST_Intersects(catchment.geometry, ST_GeomFromWKB(NULL, 4326))`. Note the `NULL` inside the `ST_GeomFromWKB` function instead of embedding the hex string for `wkb`.

I *think* the cause of this issue is that the function code is using the `WKBElement.data` attribute, which is a `buffer` type but I think it is expecting a string. Maybe the function code should be using the `WKBElement.desc` attribute instead? (e.g. [here](https://github.com/geoalchemy/geoalchemy2/blob/a7df39a9bf8798eec764a710bdfed15a7f2ac6cb/geoalchemy2/functions.py#L190)). This only effects the `WKBElement` type, as `WKTElement.data` is already a string.

If I hack the test to set the `wkb.data` attribute to `wkb.desc`, it apparently works as intended.

adrien-berchet:
Hi @jwtrhs 
Thanks for your report.
I have a few questions:
1. what is your version of `sqlalchemy`? Because upper case functions like `func.ST_INTERSECTS` could be an issue before `sqlalchemy < 1.3.4`.
2.what is `Catchment` in your query?

I tried to add the following test in `geoalchemy2/tests/test_functional.py`:
``` python
    def test_filter_from_wkb(self):
        from shapely.geometry import box
        wkb = from_shape(box(1, 1, 2, 2), srid=4326)
        query = session.query(Poi.id).filter(func.ST_INTERSECTS(Poi.__table__.c.geom, wkb))

        conn = session.connection()
        cur = conn.connection.cursor()
        compiled = query.statement.compile(conn)
        raw_sql = cur.mogrify(str(compiled), compiled.params)
        assert raw_sql == (
[...]

PR #311, opened 2021-02-09
Fix using a WKBElement in a function rendering a NULL value

jwtrhs:
Fix for #310


jwtrhs:
Think I have jumped the gun on this fix and might not be the correct way to resolve the issue, I'll close this for now.

PR #313, opened 2021-03-30
Fix CI

adrien-berchet:


adrien-berchet:
Hi @zzzeek 
I had to add this `session.expunge_all()` because of a weird faillure in https://github.com/geoalchemy/geoalchemy2/runs/2222397467?check_suite_focus=true.
After some local tests, it fails only for `py37-sqlalatest` job if the `py37-sqla11` was run just before but it doesn't fail when run alone. I don't know if it is of any importance but I prefer reporting it to you, just in case it hides a bug somewhere.

zzzeek:
that error looks to be related to garbage collection.    it might be a bug in SQLAlchemy to some extent if that's possible because we aren't supposed to assume .obj() is non-None if we know the object might not be strongly referenced.   to fix the test for now though, somewhere in there you're dealing with an ORM object of some kind, assign it to a variable first so that it isn't garbage collected when the flush happens.

adrien-berchet:
Ok, I see. Thanks for your feedback!

PR #316, opened 2021-05-25
Add support for N-D index creation

dlbrittain:
Add support for N-D index creation. See here from the [Postgis docs](https://postgis.net/workshops/postgis-intro/3d.html#n-d-indexes)

Adds an additional flag "use_N_D_index" to the Geometry options. Checks if the spatial_index is false and will return an ArgumentError.

### DIFF
--- a/geoalchemy2/__init__.py
+++ b/geoalchemy2/__init__.py
@@ -105,7 +105,7 @@ def _setup_ddl_event_listeners():
 
                         index_name = "idx_{}_{}".format(table.name, c.name)
 
-                        if c.type.is_N_D_index:
+                        if c.type.use_N_D_index:
                             gis_column = '{} gist_geometry_ops_nd'.format(c.name)
                         else:
                             gis_column = c.name
@@ -125,8 +125,8 @@ def _setup_ddl_event_listeners():
                         raise ArgumentError('dialect {} is not supported'.format(bind.dialect.name))
 
                 if isinstance(c.type, (Geometry, Geography)) and c.type.spatial_index is False and \
-                        c.type.is_N_D_index is True:
-                    raise ArgumentError('Arg Error(is_N_D_index): spatial_index must be True')
+                        c.type.use_N_D_index is True:
+                    raise ArgumentError('Arg Error(use_N_D_index): spatial_index must be True')
 
                 # Add spatial indices for the Raster columns
                 #
@@ -143,7 +143,7 @@ def _setup_ddl_event_listeners():
                                  'USING GIST (ST_ConvexHull("%s"))' %
                                  (table.name, c.name, table.name, c.name))
                     bind.execute(q)
-                
+
         elif event == 'after-drop':
             # Restore original column list including managed Geometry columns
             table.columns = table.info.pop('_saved_columns')
--- a/geoalchemy2/__init__.py
+++ b/geoalchemy2/__init__.py
@@ -94,7 +94,6 @@ def _setup_ddl_event_listeners():
                     stmt = select([func.AddGeometryColumn(*args)])
                     stmt = stmt.execution_options(autocommit=True)
                     bind.execute(stmt)
-
                 # Add spatial indices for the Geometry and Geography columns
                 if isinstance(c.type, (Geometry, Geography)) and \
                         c.type.spatial_index is True:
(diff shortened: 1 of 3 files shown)


adrien-berchet:
Hi @dlbrittain 
Thank you for this contribution!
It looks almost good to me but could you also add some tests for this new feature?

dlbrittain:
Thanks! I will add one more test to assert the error raising and I think it should be good.

dlbrittain:
Added the test, I also found test_index_without_schema was referencing the incorrect model, i fixed that as well.

geoalchemy2/__init__.py
```diff
@@ -131,7 +143,7 @@ def dispatch(event, table, bind):
                                  'USING GIST (ST_ConvexHull("%s"))' %
                                  (table.name, c.name, table.name, c.name))
                     bind.execute(q)
-
+
```
adrien-berchet, line 146: trailing spaces

geoalchemy2/types.py
```diff
@@ -129,7 +132,7 @@ class _GISType(UserDefinedType):
         geometry/geography columns. """
 
     def __init__(self, geometry_type='GEOMETRY', srid=-1, dimension=2,
-                 spatial_index=True, management=False, use_typmod=None,
+                 spatial_index=True, is_N_D_index=False, management=False, use_typmod=None,
```
adrien-berchet, line 135: What about `use_N_D_index` instead of `is_N_D_index`?

tests/test_functional.py
```diff
@@ -133,9 +139,57 @@ def test_index_with_schema(self):
...
+                                    is_N_D_index=True))
+
+        with pytest.raises(ArgumentError) as excinfo:
+            NDIndexArgErrorSchema.__table__.create(engine)
+        assert "Arg Error(is_N_D_index): spatial_index must be True" == excinfo.value.args[0]
+
     def test_index_without_schema(self):
         inspector = get_inspector(engine)
-        indices = inspector.get_indexes(IndexTestWithSchema.__tablename__)
+        indices = inspector.get_indexes(IndexTestWithoutSchema.__tablename__)
```
adrien-berchet, line 192: Good catch, thanks!

geoalchemy2/types.py
dlbrittain: I like that better, will implement

geoalchemy2/__init__.py
```diff
@@ -103,19 +102,32 @@ def dispatch(event, table, bind):
...
+                                                                             gis_column)
+
+                        q = text(sql)
+
                         bind.execute(q)
                     else:
                         raise ArgumentError('dialect {} is not supported'.format(bind.dialect.name))
 
+                if isinstance(c.type, (Geometry, Geography)) and c.type.spatial_index is False and \
+                        c.type.use_N_D_index is True:
```
adrien-berchet, line 128:
```suggestion
                if c.type.spatial_index is False and c.type.use_N_D_index is True:
```
I think that `use_N_D_index` should never be `True` when `spatial_index` is `False`, regardless of the type of `c.type`. WDYT?

geoalchemy2/__init__.py
```diff
@@ -103,19 +102,32 @@ def dispatch(event, table, bind):
                         stmt = stmt.execution_options(autocommit=True)
                         bind.execute(stmt)
                     elif bind.dialect.name == 'postgresql':
+
+                        index_name = "idx_{}_{}".format(table.name, c.name)
+
+                        if c.type.use_N_D_index:
```
adrien-berchet, line 108: I never used ND index but does it make sense to use it on 2D data? Shouldn't we check that `dimension >= 3` or something like that?

adrien-berchet (commented):
Looks good! I just have 2 last questions :)

geoalchemy2/__init__.py
dlbrittain: hmm, that is a good idea, if there is no higher dimensional data there should be no reason to index in higher dimensions!
Indeed from the [PostGIS documentation about N-D indexes ](https://postgis.net/workshops/postgis-intro/3d.html#n-d-indexes) they mention the following:

> For a higher dimension index to be useful, the data must cover a wide range of that dimension, relative to the kinds of queries you are constructing.

geoalchemy2/__init__.py
dlbrittain: Correct, this just checks that state.
A potential better way of handling this could be done under the hood and set the 'use_N_D_index' flag automatically to False rather than raising an error.

dlbrittain: Something like this?
```
if c.type.spatial_index is False:
    c.type.use_N_D_index = False
```

geoalchemy2/__init__.py
adrien-berchet: Is it possible to create such index, even though it seems useless? In geoalchemy2 we don't want to add more constraints than postgis, so if it is possible to create ND index for 2d data we can keep it like this.

dlbrittain: I can't think of a use case, but indeed there is no constraint from PostGIS as far as I know

adrien-berchet: Ok, let's not add new constraint then.

adrien-berchet:

PR #321, opened 2021-06-22
Fix quotes in index creation

### LINKED ISSUE #320: schema in uppercase ignored in version 0.9.0 (worked fine with 0.8.5)
Example:

`df.to_postgis(con=engine_postgis, name="tablename", schema="UPPERCASENAME")`

is converted to uppercasename.tablename and therefore fails as it cannot find the relevant schema (which is case sensitive). This exact code used to work in version 0.8.5 (case was kept).

Thanks for checking.

adrien-berchet:
Fixes #320


adrien-berchet:
Hello @dlbrittain 
Could you check on your side that this fix does not break N-D index introduced in your PR #316 please?

tomasjerabek-com:
Seems to work fine. Thanks @adrien-berchet :thumbsup:

PR #324, opened 2021-07-12
Add support for N-D intersects '&&&' operator

dlbrittain:
Add support for N-D spatial intersection for querying geometries in bounding boxes
See here for info:
http://postgis.net/docs/geometry_overlaps_nd.html
List of all operators in Postgis
http://postgis.net/docs/reference.html#Operators

### DIFF
--- a/geoalchemy2/comparator.py
+++ b/geoalchemy2/comparator.py
@@ -53,7 +53,7 @@ except ImportError:  # SQLA < 0.9  # pragma: no cover
 
 
 INTERSECTS = operators.custom_op('&&')
-INTERSECTS_3D = operators.custom_op('&&&')
+INTERSECTS_ND = operators.custom_op('&&&')
 OVERLAPS_OR_TO_LEFT = operators.custom_op('&<')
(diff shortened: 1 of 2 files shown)


adrien-berchet:
Hello @dlbrittain 
This looks nice, thank you very much for this!

geoalchemy2/comparator.py
```diff
@@ -220,3 +221,10 @@ def distance_box(self, other):
         geometries.
         """
         return self.operate(DISTANCE_BOX, other, result_type=DOUBLE_PRECISION)
+
+    def intersects_3d(self, other):
```
adrien-berchet, line 225: I am just wondering about the name, shouldn't it be `intersects_nd`?

adrien-berchet: :+1:

adrien-berchet:
Thank you very much!
Do you need a new release?

dlbrittain:
It would be very helpful indeed! Thank you!

adrien-berchet:
The version 0.9.2 was released :)

PR #328, opened 2021-07-20
Bigquery

jimfulton:
This PR adds support for BigQuery. (Changes will also be needed in in the BigQuery SQLAlchemy dialect.) It adds some special cases similar to those for sqlite.

I could add geography support without these change, but that would require [subclassing geoalchemy2 classes](https://github.com/jimfulton/python-bigquery-sqlalchemy/blob/geoalchemy/sqlalchemy_bigquery/geography.py), which I'd prefer not to do.

BigQuery only supports one projection, WGS84, SRID 4326, and functions that accept `SRID` arguments in PostGIS don't in BigQuery.  Also, BigQuery doesn't support geometry, or have `ST_GEOM...' functions.

I'd like to get a read on whether the approach is acceptable. If it is, then I'll add tests and finish the BigQuery change before making this a non-draft PR.

### DIFF
--- a/geoalchemy2/__init__.py
+++ b/geoalchemy2/__init__.py
@@ -123,6 +123,8 @@ def _setup_ddl_event_listeners():
                         q = text(sql)
 
                         bind.execute(q)
+                    elif bind.dialect.name == 'bigquery':
+                        pass  # No indexes for BigQuery
                     else:
                         raise ArgumentError('dialect {} is not supported'.format(bind.dialect.name))
 
--- a/geoalchemy2/types.py
+++ b/geoalchemy2/types.py
@@ -123,6 +123,10 @@ class _GISType(UserDefinedType):
     """ The name of "from text" function for this type.
         Set in subclasses. """
 
+    from_binary = None
+    """ The name of "from binary" function for this type.
+        Set in subclasses. """
+
     as_binary = None
     """ The name of the "as binary" function for this type.
         Set in subclasses. """
@@ -161,6 +165,11 @@ class _GISType(UserDefinedType):
     def result_processor(self, dialect, coltype):
         def process(value):
             if value is not None:
+                if dialect.name == 'bigquery':
+                    element = WKBElement(value, srid=4326, extended=True)
+                    element.geom_from_extended_version = 'ST_GeogFromWKB'
+                    return element
+
                 kwargs = {}
                 if self.srid > 0:
                     kwargs['srid'] = self.srid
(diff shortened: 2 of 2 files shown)


jimfulton:
It turns out that the main advantage I was hoping for by not subclassing doesn't exist.

Still, the subclassing approach hides geoalchemy2 somewhat, which is a disadvantage, IMO.

OTOH, subclassing makes this and changes to geoalchemy2 unnecessary.

I'm interested in your thoughts. :)

adrien-berchet:
Hello @jimfulton 
Thank you very much for this work. I will try to take some time to think about this next week. At first glance I looks quite good to me.
I think it would be much simpler for users if GeoAlchemy2 can support BigQuery natively so they don't have to subclass anything.

jimfulton:
> Hello @jimfulton
> Thank you very much for this work. I will try to take some time to think about this next week. At first glance I looks quite good to me.
> I think it would be much simpler for users if GeoAlchemy2 can support BigQuery natively so they don't have to subclass anything.

Users wouldn't have to subclass anything. BigQuery would: https://github.com/jimfulton/python-bigquery-sqlalchemy/blob/geoalchemy/sqlalchemy_bigquery/geography.py.

And even with the PR here, changes are needed to the BigQuery SQLAlchemy dialect.

The main difference for users is whether they:

`from geoalchemy2 import Geography'

or

`from sqlalchemy-bigquery import GEOGRAPHY` (where GEOGRAPHY is a subclass of Geography)

Also, this PR is incomplete. It needs tests.  I wanted to discuss the approach first. If you're good with the approach, then I'll add tests.

jimfulton:
Looking at the testing situation in geoalchemy2, I think it would be better to do everything on the BigQuery side.

The way your testing is set up, you'd end up making BigQuery connections in your tests, which would be a pretty big hassle.

I could write connection-less tests for some, but not all of the changes I made.

Since I have to make changes on the BigQuery side anyway, I think it makes more sense to make all of the changes there.  I can do all of the changes on our side.

adrien-berchet:
> > I think it would be much simpler for users if GeoAlchemy2 can support BigQuery natively so they don't have to subclass anything.
> 
> Users wouldn't have to subclass anything. BigQuery would: https://github.com/jimfulton/python-bigquery-sqlalchemy/blob/geoalchemy/sqlalchemy_bigquery/geography.py.
> 
> And even with the PR here, changes are needed to the BigQuery SQLAlchemy dialect.
> 
> The main difference for users is whether they:
> 
> `from geoalchemy2 import Geography`
> 
> or
> 
> `from sqlalchemy-bigquery import GEOGRAPHY` (where GEOGRAPHY is a subclass of Geography)

Ah ok I see.
Indeed, after reading your code I think it makes more sense to subclass it on you side, as the Geography objects of BigQuery are quite specific. And for the user it seems more convenient to do `from sqlalchemy-bigquery import GEOGRAPHY`.

PR #330, opened 2021-07-29
WIP: Add MySQL support

### LINKED ISSUE #318: Support for MySQL spatial queries
Hello, I am a new to the Geoalchemy community and just starting to understand spatial data types and querying. 

I noticed that Geoalchemy focused on PostGIS, and I wanted to use this issue to understand if there are plans to support MySQL spatial extensions in the future.

1. Are there plans to support MySQL in the future? Are there other packages that do this already?
2. Are there challenges to supporting MySQL spatial extensions that I should be aware of?
3. I am interested in contributing to the project by adding MySQL support, depending on 1 & 2  — would that be a useful addition to this project?

Look forward to learning more!

adrien-berchet:
Fixes #318 

This is an attempt to add MySQL support.


adrien-berchet:
The geometry columns and spatial indexes seems to be properly created.
The next issue is that (as far as I can see in the MySQL doc) we need to pass the SRID to the function `ST_GeomFromText` when data are inserted and the column has an SRID, which is currently not possible with `geoalchemy2`.

adrien-berchet:
@krisHans3n I rebased this branch in case you would like to work on it :-)

krisHans3n:
> @krisHans3n I rebased this branch in case you would like to work on it :-)

That's great, thank you. Apologies for not having started on this yet, I will get round to it as soon as I possibly can

adrien-berchet:
> > @krisHans3n I rebased this branch in case you would like to work on it :-)
> 
> That's great, thank you. Apologies for not having started on this yet, I will get round to it as soon as I possibly can

No worries, this is not supposed be part of the next release, we have time to work on this.
Thank you anyway to take some time on this project! :-)

adrien-berchet:
Hi @zzzeek 

I have an issue that I don't know how to solve... so I have to call for your help :smiley: 

The custom types defined in `GeoAlchemy2` automatically add SQL expressions when inserting or querying a Geometry column. For this the methods `column_expression`, `result_processor`, `bind_expression` and `bind_processor` are overridden (see https://github.com/geoalchemy/geoalchemy2/blob/fd3fce052bdc734aea42dac66bd7fc4603ccbbbd/geoalchemy2/types.py#L166). With these methods we add `ST_GeomFromEWKT` / `ST_AsBinary` to insert / select expressions. But now I would like to add support for MySQL and it's becoming tricky. First of all, MySQL does not know `ST_GeomFromEWKT` but only `ST_GeomFromText`, which is possible to replace using the `@compiles` decorator. But what I don't know how to solve is how to split the parameters, because for other supported dialects we call something like `ST_GeomFromEWKT('SRID=4326;POINT(0 1)')` while for MySQL we should call `ST_GeomFromText('POINT(0 1)', 4326)`. So I have to split the parameter into 2, and I don't know how to solve this with SQLAlchemy. I tried several things with the `@compiles` decorator but there is always something missing to make it work. If you have any hint it would be super nice.

Thanks!

zzzeek:
have never had to support that before.   with caching, it's weird.

option 1: might be best to record three bind parameters up front, one is the "joined" one and the other two are separate.  then have the two different compiler schemes use the one param, or the two separate ones.   that way you are capturing all three params so that the construct's SQL generation can still be cached.   

option 2, this is likely better: rework the original one to use two bound parameters and use string concatenation in the rendered SQL.

adrien-berchet:
Thanks for your answer!
I will try to work on that soon. 

But just to be sure I understand your suggestion, for the option 1 you mean that
```python
    def bind_expression(self, bindvalue):
        return getattr(func, self.from_text)(bindvalue, type_=self)
```
would become
```python
    def bind_expression(self, bindvalue):
        return getattr(func, self.from_text)(bindvalue, bindvalue.data, bindvalue.srid, type_=self)
```
right? But then how can I define the 'two different compiler schemes'?

And for option 2,
```python
    def bind_expression(self, bindvalue):
        return getattr(func, self.from_text)(bindvalue, type_=self)
```
would become
```python
    def bind_expression(self, bindvalue):
        return getattr(func, self.from_text)(bindvalue.data, bindvalue.srid, type_=self)
```
right? And then I can use the same compiler scheme for all dialects. But my issue in this case is that for extended spatial elements (EWKT and EWKB), the SRID is already contained in the data, so I would have to discard it before compiling the SQL, which is a bit annoying.

zzzeek:
lets stick with option 2, you just redefine PG's function to use two bound parameters.

I dont know what datatype is returned by `getattr(func, self.from_text)` but something that you can run through `@compiles`, sure

adrien-berchet:
Hi @zzzeek 
Thanks again for your answer, I finally found a way to make it work.

In the end, I kept the `bind_expression` as it was (so the SRID is still not passed at this point), the `bindvalues` are processed in `bind_processor` in order to be compatible with the given function (which I can retrieve from `self.from_text`), and then the `@compiles` step adds the SRID of the column to the function (this is ok since we check the SRID in the `bind_processor` function and raise an exception if it is different from the one of the column).

I did this because I had the following issues, especially because of the `executemany` mode:

* In `bind_expression` it is not possible to extract the SRID of the given value in `executemany` mode. Also, we can't know the dialect used in this function, which is an issue for me because I would like to keep the other dialects as they are for performance reasons.
* In `@compiles` it is not possible to get the actual value in `executemany` mode (we just get a `symbol(REQUIRED)`), so it is not possible to fix the SRID at this point.
* The MySQL dialect handles the spatial data quite differently from PostgreSQL and SQLite. For example it does not support extended WKT/WKB, and the input functions behaves a bit differently, so it's hard anyway to build a process that works as well as for other dialects.

But now I think we are able to support most of the simple cases, so it's already a big step for GeoAlchemy2.

Thanks again!

adrien-berchet:
@marecabo @giuliano-oliveira @krisHans3n @jakubczaplicki @hemna @rtindru
I saw that you seem interested by this feature, could you take some time to test this branch on your real use cases please? It would help me check that it works fine. Thanks!

adrien-berchet:
> @marecabo @giuliano-oliveira @krisHans3n @jakubczaplicki @hemna @rtindru I saw that you seem interested by this feature, could you take some time to test this branch on your real use cases please? It would help me check that it works fine. Thanks!

I wanted to work on something else so I merged. Nevertheless, if any of you can test the MySQL dialect on the master branch before the next release I would be grateful.

giuliano-oliveira:
@adrien-berchet I would happily do, but I will probably have time for this next weekend, is this time ok?
Do we have a estimate for when the next release would be deployed?

adrien-berchet:
Hi @giuliano-oliveira 
Thank you very much! There is no schedule for the next release so take your time :⁠-⁠)

PR #331, opened 2021-08-25
Fix warnings in preparation for SQLAlchemy v2.0

robintw:
We are currently in the process of updating our application in preparation for SQLAlchemy v2.0. The current release of SQLAlchemy is v1.4, which has an environment variable setting (`SQLALCHEMY_WARN_20=1`) which will enable warnings for functionality that will be removed in SQLAlchemy v2.0.

I found that a couple of lines of code in geoalchemy2 emitted a warning in this situation.

```
  /Users/robin/anaconda3/envs/pepys_sa14/lib/python3.8/site-packages/geoalchemy2/__init__.py:96: RemovedIn20Warning: The legacy calling style of select() is deprecated and will be removed in SQLAlchemy 2.0.  Please use the new calling style described at select(). (Background on SQLAlchemy 2.0 at: https://sqlalche.me/e/b8d9)
    stmt = select([func.AddGeometryColumn(*args)])
```

```
  /Users/robin/anaconda3/envs/pepys_sa14/lib/python3.8/site-packages/geoalchemy2/__init__.py:104: RemovedIn20Warning: The legacy calling style of select() is deprecated and will be removed in SQLAlchemy 2.0.  Please use the new calling style described at select(). (Background on SQLAlchemy 2.0 at: https://sqlalche.me/e/b8d9)
    stmt = select([func.CreateSpatialIndex(table.name, c.name)])
```

Fixing this is really easy - as it just involves removing the square brackets surrounding the function call.

This fix works fine in SQLAlchemy versions prior to 1.4 as well.

### DIFF
--- a/geoalchemy2/__init__.py
+++ b/geoalchemy2/__init__.py
@@ -93,7 +93,7 @@ def _setup_ddl_event_listeners():
                     if bind.dialect.name == 'sqlite':
                         args.append(not c.type.nullable)
 
-                    stmt = select([func.AddGeometryColumn(*args)])
+                    stmt = select(func.AddGeometryColumn(*args))
                     stmt = stmt.execution_options(autocommit=True)
                     bind.execute(stmt)
 
@@ -101,7 +101,7 @@ def _setup_ddl_event_listeners():
                 if isinstance(c.type, (Geometry, Geography)) and \
                         c.type.spatial_index is True:
                     if bind.dialect.name == 'sqlite':
-                        stmt = select([func.CreateSpatialIndex(table.name, c.name)])
+                        stmt = select(func.CreateSpatialIndex(table.name, c.name))
                         stmt = stmt.execution_options(autocommit=True)
                         bind.execute(stmt)
                     elif bind.dialect.name == 'postgresql':
--- a/geoalchemy2/__init__.py
+++ b/geoalchemy2/__init__.py
@@ -15,9 +15,13 @@ from .exc import ArgumentError
 from . import functions  # NOQA
 from . import types  # NOQA
 
+import sqlalchemy
 from sqlalchemy import Table, event
 from sqlalchemy.sql import select, func, expression, text
 
+from packaging import version
+
+SQLALCHEMY_VERSION_BEFORE_14 = version.parse(sqlalchemy.__version__) < version.parse("1.4")
 
 def _setup_ddl_event_listeners():
     @event.listens_for(Table, "before_create")
@@ -69,7 +73,10 @@ def _setup_ddl_event_listeners():
                     args = [table.schema] if table.schema else []
                     args.extend([table.name, c.name])
 
-                    stmt = select(getattr(func, drop_func)(*args))
+                    if SQLALCHEMY_VERSION_BEFORE_14:
+                        stmt = select([getattr(func, drop_func)(*args)])
+                    else:
+                        stmt = select(getattr(func, drop_func)(*args))
                     stmt = stmt.execution_options(autocommit=True)
                     bind.execute(stmt)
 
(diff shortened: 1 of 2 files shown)


adrien-berchet:
Hello @robintw 
Thank you very much for this, I did not try SQLAlchemy2 yet so I don't know what should be changed to be compatible.

adrien-berchet:
Ah we might have to increase the min version of the SQLAlchemy dependency. Do you know if it is possible to make a change that would be compatible with both SQLAlchemy == 1.1 and 2?

robintw:
I've just seen the error in CI myself and was wondering the same thing.

From what I can see from a look around the docs, I don't think it is possible to do it in a way that works for 1.1 and 2.0. If you have the square brackets then 1.1 works and 2.0 will fail, and if you don't have the square brackets then 1.1 will fail.

I think probably the easiest way is to up the dependency to 1.2 - would you be ok with that?

adrien-berchet:
I think we will have to increase it anyway. I would just like to avoid increasing it too much, but if it works with 1.2 it's good.

robintw:
I haven't had chance to test with 1.2 yet - I'll try and do a test locally soon

robintw:
Unfortunately it seems that the new way of passing arguments to `select()` is only available from v1.4. I'll see if I can put in some sort of version check or something.

robintw:
@adrien-berchet I think I've fixed this now with a version check - but the CI seems to require your permission to run (I've tested it locally, but also come across a load of postgres errors that I think are to do with my setup not being quite the same as the CI)

geoalchemy2/__init__.py
```diff
@@ -15,9 +15,13 @@
 from . import functions  # NOQA
 from . import types  # NOQA
 
+import sqlalchemy
 from sqlalchemy import Table, event
 from sqlalchemy.sql import select, func, expression, text
 
+from packaging import version
+
+SQLALCHEMY_VERSION_BEFORE_14 = version.parse(sqlalchemy.__version__) < version.parse("1.4")
```
adrien-berchet, line 24:
```suggestion
SQLALCHEMY_VERSION_BEFORE_14 = version.parse(sqlalchemy.__version__) < version.parse("1.4")

def _format_select_args(*args):
    if _SQLALCHEMY_VERSION_BEFORE_14:
        return [args]
    else:
        return args
```
I think we could use a function like this to format args.
And then we could replace
```python
                    if SQLALCHEMY_VERSION_BEFORE_14:
                        stmt = select([getattr(func, drop_func)(*args)])
                    else:
                        stmt = select(getattr(func, drop_func)(*args))
```
 by
```python
                    stmt = select(*_format_select_args(getattr(func, drop_func)(*args)))
```

adrien-berchet (commented):
Appart to the lint error it looks good to me. I just have a suggestion to make the code less verbose.

adrien-berchet:
You no longer have any `RemovedIn20Warning` with this change? That's all we have to do to support `SQLAlchemy >= 2`? That was easier than I thought :-)

robintw:
I no longer have any RemovedIn20Warnings when using geoalchemy with my application code. I think there will probably be some bits of the geoalchemy test suite itself that need changing - from a brief look, I think there is the same select with a list vs positional arguments issue there - but it's hard for me to fix those as I can't seem to get the test suite running properly with Postgres on my machine.

Basically, I think this is done in terms of supporting other people running SQLAlchemy 2.0, but needs a bit more to make it support actually running the geoalchemy2 test suite under SQLAlchemy 2.0.

adrien-berchet:
Ah yes indeed, I started to update a few tests to make them work with SQLA>=2 but I will fix this later, when I can actually test with SQLA2 to be sure everything works. So let's merge this one for now.

robintw:
That's wonderful - thank you.

How often do you tend to do releases? It'd be great if a release could be done soon, so that we can install directly from pip rather than through git.

adrien-berchet:
Well, there is no usual release plan, we just release when there is something new. I can do a new one.

PR #338, opened 2021-12-13
Fix cache warnings

adrien-berchet:


adrien-berchet:
Hi @zzzeek 
Sorry I need you again :)
I am fixing the new warnings for the cache issues but I am not sure in which case we should use the cache or not and I don't have much time to investigate this in details... In this PR I assumed that only the objects whose internal state does not change the SQL string can use the cache (which is unfortunately not always the case for some classes provided by GeoAlchemy2), but I am not sure.
I disabled the cache for the types and enabled it for the functions (enabling it for the types break a test for an unknown reason so I disabled it).
Could you take a quick look on this please?

geoalchemy2/elements.py
```diff
@@ -253,6 +253,8 @@ class CompositeElement(FunctionElement):
     Instances of this class wrap a Postgres composite type.
     """
 
+    inherit_cache = True
+
```
zzzeek, line 257: if I have two CompositeElement objects with a different "field" and "type_", is the identical SQL produced for the two objects?  is the return type of the SQL function exactly the same for both objects? if no, this has to be inherit_cache=False unless a cache key that takes these fields into account is implemented (which as noted, I haven't made that API very public yet).

geoalchemy2/functions.py
```diff
@@ -75,6 +75,9 @@ def __init__(cls, clsname, bases, clsdict):
 
 
 class TableRowElement(ColumnElement):
+
+    inherit_cache = True
+
```
zzzeek, line 80: this one also looks like inherit_cache=False

if I have:

t1 = TableRowElement(select(Foo))
t2 = TableRowElement(select(Bar))

do t1 / t2 produce identical SQL ?   keep in mind, inherit cache means, t1 and t2 will have the same cache key.  t1 will generate SQL, t2 will not.    seems unlikely here

geoalchemy2/functions.py
```diff
@@ -88,6 +91,7 @@ class ST_AsGeoJSON(with_metaclass(_GenericMeta, functions.GenericFunction)):
     feature version introduced in PostGIS 3."""
 
     name = "ST_AsGeoJSON"
+    inherit_cache = True
```
zzzeek, line 94: this one might be OK since there's no additional state here that isnt derived from the arguments that are already part of the cache key, but if you are in a hurry and dont have time to test very carefully, I'd keep inherit_cache=False

tests/gallery/test_summarystatsagg.py
```diff
@@ -34,12 +34,16 @@ class SummaryStatsCustomType(CompositeType):
...
 
+    cache_ok = True
+
 
 class ST_SummaryStatsAgg(GenericFunction):
     type = SummaryStatsCustomType
     # Set a specific identifier to not override the actual ST_SummaryStatsAgg function
     identifier = "ST_SummaryStatsAgg_custom"
 
+    inherit_cache = True
```
zzzeek, line 45: should be OK

geoalchemy2/types.py
```diff
@@ -403,3 +418,6 @@ class SummaryStats(CompositeType):
         'min': Float,
         'max': Float,
     }
+
+    cache_ok = True
```
zzzeek, line 422: should be OK

geoalchemy2/types.py
```diff
@@ -385,6 +397,9 @@ class GeometryDump(CompositeType):
     typemap = {'path': postgresql.ARRAY(Integer), 'geom': Geometry}
     """ Dictionary defining the contents of a ``geometry_dump``. """
 
+    cache_ok = True
```
zzzeek, line 400: should be OK

zzzeek (changes requested):
heya

if you dont have time to test these constructs very carefully in a caching scenario, I would advise setting all the inherit_cache to False for now; most of them look not ready for caching yet in any case as they contain state that seems likely to affect SQL generation or return types.

ideally for caching you'd have some SQL statements that are run many times, given a sample of the construct where you change it on each run, and ensure the correct SQL / parameters are emitted.

geoalchemy2/elements.py
adrien-berchet: Ok, I see, let's set it to False then.

geoalchemy2/functions.py
adrien-berchet: Indeed, I think they could produce different SQL so let's set it to False.

geoalchemy2/functions.py
adrien-berchet: I think the SQL will only depend on the arguments so I guess they should be part of the cache key. Let's try to keep True for this one.

adrien-berchet:
Thank you very much for these insights @zzzeek !
I am still not sure to understand how the cache keys are constructed so it was hard to choose which objects should use the cache or not. For now I am going to disable the cache when I am not sure and we'll see if it causes perf issues.
Thanks again!

zzzeek:
> Thank you very much for these insights @zzzeek ! I am still not sure to understand how the cache keys are constructed so it was hard to choose which objects should use the cache or not. For now I am going to disable the cache when I am not sure and we'll see if it causes perf issues. Thanks again!

that's fine, you wont see perf issues compared to 1.3.x except in the area where applications use a lot of ORM lazy loading and your special functions happen to be in those queries, those will not be cached whereas in 1.3 they were.

re: the caching I would say there's no need to rush turning it on, when you have time I can try to give you more background on what exactly it does.   I've worked with these caching approaches for years now but it should be expected they can easily seem opaque at first.

PR #343, opened 2022-01-05
Fix creation of columns using a TypeDecorator

### LINKED ISSUE #342: Question: Is there a best practices way to support TypeDecorator classes?
I currently use two database backends for a project that I'm working on: postgresql and sqlite. I've been attempting to use geoalchemy2 to help add geospatial support to my application. I would like to design a custom type for a geospatial column using sqlalchemy's `TypeDecorator` methodology, where the implementation is based on geoalchemy2's `Geometry` class, but that can specify a `Geography` class as column type in the case where I'm connecting to or creating a geospatial column in a table with postgis.

Taking a look at the code, I see that geoalchemy2's `__init__.py` sets up a series of event handlers for managing addition of `Geometry` columns in a database. However, it requires the Column to be specified as a `Geometry` column specifically in order to correctly handle the geospatial column (in the case of sqlite), and fails to catch the custom typed class column as being a geospatial column.

One idea that I had to allow my `TypeDecorator` custom type class to be "seen" as a GIS type column is to essentially remove the events that are created in geoalchemy2's `__init__.py`, copy the event setup functions to my own script, add functionality so that the custom type class is seen as GIS type, then re-run the event handlers. However, this feels brittle and unsustainable in the long run. 

Is there a better method that would be recommended for a case like this? Maybe something using the custom type class itself?

An example of the sort of custom type class I would like to use is shown below.

```python
import sqlalchemy as sa
from geoalchemy2.types import Geography, Geometry
from geoalchemy2.shape import to_shape
from sqlalchemy.ext.declarative import declarative_base

class HybridGeospatial(sa.types.TypeDecorator):
    """
    Column type for Geospatial data points in geodetic point coordinates (longitude, latitude).
    """
    impl = Geometry
    
    def load_dialect_impl(self, dialect):
        if dialect.name == 'postgresql':
            return dialect.type_descriptor(Geography)
        else:
            return dialect.type_descriptor(Geometry)

    def process_bind_param(self, coords, dialect):
        """
        Take a tuple or 2-D array `coords` representing lon/lat coordinates, and return WKT
        """
        if coords is None:
            return coords
        if len(coords) != 2:
            raise TypeError("Incorrect length for bound parameter of type HybridGeospatial")

        return f"POINT ({coords[0]} {coords[1]})"

    def process_result_value(self, result_wkb, dialect):
        """
        Result value will be in well-known binary format
        """
        if result_wkb is None:
            return result_wkb

        return to_shape(result_wkb).centroid.coords[0]


Base = declarative_base()

class GeoTable(Base):
    __tablename__ = 'geo_table'
    id = sa.Column(sa.Integer, primary_key=True)
    geom = sa.Column(HybridGeospatial(geometry_type='POINT', management=True, dimension=2))

GeoTable.__table__.create(engine)
```

Thanks in advance for any help!

adrien-berchet:
Fix #342

### DIFF
--- a/geoalchemy2/__init__.py
+++ b/geoalchemy2/__init__.py
@@ -18,6 +18,7 @@ from . import types  # NOQA
 import sqlalchemy
 from sqlalchemy import Table, event
 from sqlalchemy.sql import select, func, expression, text
+from sqlalchemy.types import TypeDecorator
 
 from packaging import version
 
@@ -29,6 +30,16 @@ def _format_select_args(*args):
         return [args]
     else:
         return args
+
+
+def _check_spatial_type(tested_type, spatial_types):
+    return (
+        isinstance(tested_type, spatial_types)
+        or (
+            isinstance(tested_type, TypeDecorator)
+            and isinstance(tested_type.impl, spatial_types)
+        )
+    )
 
 
 def _setup_ddl_event_listeners():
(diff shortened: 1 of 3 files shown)


ryangooch:
I tested this branch on my machine, using the `TypeDecorator` class defined in #342 , and it appears to work perfectly/exactly as intended. Thanks!

PR #344, opened 2022-01-05
Fix spatial index declaration

adrien-berchet:


spolloni:
@adrien-berchet sorry a little late to the party, but thanks for this.

adrien-berchet:
> @adrien-berchet sorry a little late to the party, but thanks for this.

No problem, you're welcome :-)

spolloni:
@adrien-berchet for posterity would you mind taking a few minutes to add a description of what exactly this PR implements? We just bumped geoalchemy in one of our apps, and this caused a few VIEW-related bugs that took me while to understand, mostly because my memory was quite foggy, but also because there wasn't much to read here.

adrien-berchet:
Hi @spolloni 
I just updated the description, is it clear enough?
You raise an interesting issue about views, I never tested this. Maybe we could add a few tests with views and try to detect this case to not try to attach an `Index` to views. I don't know if SQLAlchemy is able to make the difference between a table and a view?

spolloni:
> is it clear enough?

Yes thanks a bunch!

> Maybe we could add a few tests with views and try to detect this case to not try to attach an `Index` to views. I don't know if SQLAlchemy is able to make the difference between a table and a view?

Good idea, and off the top of my head I don't know either.  I'll try to look into it but the at least the postgres errors I was getting made it pretty clear what was going on.

adrien-berchet:
Ok, thanks :)
I added a simple test in #391 with a note for later. I didn't find a simple way to check if a `Table` object refers to a view, but I didn't took much time for this.

PR #362, opened 2022-02-28
Add Alembic helpers for Add/Drop columns with SQLite

### LINKED ISSUE #351: Adding SpatiaLite Geometry Columns with Alembic?
I encountered this issue when trying to update existing tables to include geospatial columns. If we assume that we have some model for a class as such:

```python
class GeoTable(Base):
    __tablename__ = 'geo_table'
    id = sa.Column(sa.Integer, primary_key=True)
    foo = sa.Column(sa.Float)
```

and we want to add a `Geometry` column to this existing table in an existing database through a migration, we might have an Alembic migration version script like this

```python
[...]
def upgrade():
    op.add_column("geo_table", sa.Column("geom", Geometry("POINT", srid=4326)))

[...]
```

the migration produces the following error upon attempted migration:
```shell
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) near "POINT": syntax error
[SQL: ALTER TABLE geo_table ADD COLUMN geom geometry(POINT,4326)]
```

I suspect this is a similar issue to that which was addressed in #342, whereby a `sqlalchemy` event handler was needed in order to monitor for changes coming from a `TypeDecorator` class. 

One workaround that occurred to me was to have the migration check the dialect prior to attempting to add the column, then use the `SpatiaLite` function for adding a `Geometry` column directly. However, I'm not sure what to do in the case the column we wish to add to an existing table is defined using the `TypeDecorator` construct.

As always, I appreciate any help, and thanks again for all your work on this fantastic project!

adrien-berchet:
Fixes #351

### DIFF
--- a/doc/alembic.rst
+++ b/doc/alembic.rst
@@ -204,3 +204,77 @@ in ``my_package.custom_types``, you just have to edit the ``env.py`` file like t
         # ...
 
 Then the proper imports will be automatically added in the migration scripts.
+
+
+Add / Drop columns
+------------------
+
+Some dialects (like SQLite) require some specific management to alter columns of a table. In this
+case, other dedicated helpers are provided to handle this. For example, if one wants to add and drop
+columns in a SQLite database, the ``env.py`` file should look like the following:
+
+.. code-block:: python
+
+from alembic.autogenerate import rewriter
+
+writer = rewriter.Rewriter()
+
+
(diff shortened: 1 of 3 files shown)


ryangooch (approved):
I've taken a couple looks over this now, looks good to me! :+1:

adrien-berchet:
Cool, thank you very much!

PR #374, opened 2022-03-13
Improve reflection mechanism and alembic interactions

### LINKED ISSUE #376: Issue with PostGIS spatial index using 0.11
Hi,

we're getting some reports in the geopandas repository that updating geoalchemy2 to 0.11 breaks the ability to write GeoPandas GeoDataFrame to PostGIS. See https://github.com/geopandas/geopandas/issues/2375. I've noticed some discussion that may be related in #372 but that refers to SpatiaLite. 

I am not sure if this is an intended change of the behaviour but I am also not entirely sure what was the intended change as the PR that probably changed this (#344) has no description.

I am happy to fix that on geopandas side if you don't see it as a regression though I have to admit I am not fully sure how at this moment. We currently do not control spatial index on the database side in any way.

adrien-berchet:
# Goals of this PR:

* keep the spatial index creation easy (with `spatial_index=True`) and compatible with all SQLAlchemy mechanisms and with Alembic.
* fix the reflection mechanism for spatial types and spatial indexes.
* fix #376 

# Main issues:
* spatial indexes are properly reflected with Postgres but not with SQLIte (because they are tables).
* indexes are reflected after the columns so it is not possible to know during column processing if we should create an index or let it be reflected afterwards.
* using `_column_flag=True` makes it work with `to_metadata` but it's a private attribute so it might not be reliable?
* Alembic triggers the `before_create` and `after_create` events but not the `before_drop` and `after_drop` events on which we rely on.


adrien-berchet:
Hi @zzzeek 
Here is the following of https://github.com/sqlalchemy/alembic/issues/1034.
As I said, I still have (at least) one issue with Alembic: the `CreateTableOp` operation triggers the `before_create` and `after_create` events as expected but the `DropTableOp` does not trigger the `before_drop` and `after_drop` events on which we rely on to properly drop the spatial columns and to clean the remaining indexes with SQLite. Is there any reason for that or was it forgotten? I can create an issue in Alembic for this.

zzzeek:
the before/after create dispatch was added to support some ENUM use cases which relied upon these to create their constraints.  it looks like the current scope of before/after create/drop is limited to just MetaData and Table in SQLAlchemy (like, not for constraints or anything like that) so sure if you want to send a PR w/ tests for the drop events, that's fine, look right here: 

https://github.com/sqlalchemy/alembic/blob/188dd8b2f098e804c8e64956e9c6490d41f1d7ce/alembic/ddl/impl.py#L350-L374

PR #375, opened 2022-03-18
Functions returning Box types now return None types instead of Geometry types

EdwardBetts:
Version 0.9.0 broke my code. My query was returning a string from Box2D(), since version 0.9.0 the call to Box2D() gets wrapped by ST_AsEWKB. This change fixes my problem so I can get a string back from Box2D().

### DIFF
--- a/geoalchemy2/_functions.py
+++ b/geoalchemy2/_functions.py
@@ -473,7 +473,7 @@ _FUNCTIONS = [
      '''Window function that returns a cluster id for each input geometry using the K-means algorithm.'''),
     ('ST_ClusterWithin', types.Geometry,
      '''Aggregate function that clusters the input geometries by separation distance.'''),
-    ('Box2D', types.Geometry,
+    ('Box2D', None,
      ('''Returns a BOX2D representing the 2D extent of the geometry.''', 'Box2D_type')),
     ('Box3D', types.Geometry,
      ('''[geometry] Returns a BOX3D representing the 3D extent of the geometry.\nOR\n[raster] Returns the box 3d representation of the enclosing box of the raster.''', 'Box3D_type')),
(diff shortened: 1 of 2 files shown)


EdwardBetts:
``` python
from geoalchemy2 import Geometry
from sqlalchemy.orm import scoped_session, sessionmaker
from sqlalchemy import func

session = scoped_session(sessionmaker())

stmt = session.query(func.ST_Dump(Geometry()).label("x")).subquery()

print(session.query(func.Box2D(stmt.c.x.geom)))
```

adrien-berchet:
Hi @EdwardBetts 
Thanks for this, it's indeed probably better for now. Maybe I will add a Box type later but for now let's not wrap it.
Could you just also remove the wrapping for Box3D please?

geoalchemy2/_functions.py
```diff
@@ -473,7 +473,7 @@
      '''Window function that returns a cluster id for each input geometry using the K-means algorithm.'''),
     ('ST_ClusterWithin', types.Geometry,
      '''Aggregate function that clusters the input geometries by separation distance.'''),
-    ('Box2D', types.Geometry,
+    ('Box2D', None,
      ('''Returns a BOX2D representing the 2D extent of the geometry.''', 'Box2D_type')),
     ('Box3D', types.Geometry,
```
adrien-berchet, line 478:
```suggestion
    ('Box3D', None,
```

adrien-berchet (changes requested):

EdwardBetts:
Will do

adrien-berchet:
Thanks @EdwardBetts , I updated a few other functions for consistency, does it looks good to you?

adrien-berchet (approved):

EdwardBetts:
Looks good to me.

adrien-berchet:
Thanks!

PR #392, opened 2022-07-27
Replace pkg_resources usage with packaging + importlib.metadata

jacob-indigo:
`geoalchemy2` is currently fairly slow to import (with and without `__pycache__` being built) and, more importantly, it seems gets substantially slower as the Python environment gets larger. **This is not an issue with `geoalchemy2` itself though**. `SQLAlchemy` is the main culprit for the slow baseline import time and `pkg_resources` is the culprit for the environment size-dependent import time (see: https://github.com/pypa/setuptools/issues/926). 

The PR does two things:
1. Replaces the use of `pkg_resources` in `shape.py` with `packaging` (an existing dependency)
2. Attempts to use `importlib.metadata` instead of `pkg_resources` in `__init__.py`, falling back to `pkg_resources` if you're using Python <3.8

Profiling import time before and after these changes:
```
Tiny (43M):
    Before: 0.332
    After: 0.252

Medium (383M):
    Before: 0.419
    After: 0.319

Large (478M):
    Before: 0.429
    After: 0.308

X-Large (600M):
    Before: 0.462
    After: 0.316 (1.5x speedup)
```
The profiling was done in 4 different `virtualenv`s, covering a range of environment sizes (arbitrarily named here). Tiny is just what's in `requirement-rtd.txt` plus `shapely`: [requirements.tiny.txt](https://github.com/geoalchemy/geoalchemy2/files/9200680/requirements.tiny.txt), and the other environments are real example requirements files from internal projects (not included here but available upon request).

These times were produced simply using: 
```bash
time python -c "import geoalchemy2; import geoalchemy2.shape;"
```

Additionally, here are some visualizations of the `python -X importtime` outputs for the same profiling command in the X-Large environment:
Before:
Screen Shot 2022-07-27 at 11 50 16 AM

After:
Screen Shot 2022-07-27 at 11 50 06 AM

### DIFF
--- a/geoalchemy2/__init__.py
+++ b/geoalchemy2/__init__.py
@@ -451,15 +451,9 @@ try:
     import importlib.metadata
     __version__ = importlib.metadata.version('GeoAlchemy2')
 except ImportError:
-    pass
-
-try:
-    if __version__ == "UNKNOWN VERSION":
+    try:
         from pkg_resources import DistributionNotFound
         from pkg_resources import get_distribution
-        try:
-            __version__ = get_distribution('GeoAlchemy2').version
-        except DistributionNotFound:  # pragma: no cover
-            pass  # pragma: no cover
-except ImportError:  # pragma: no cover
-    pass  # pragma: no cover
+        __version__ = get_distribution('GeoAlchemy2').version
+    except (DistributionNotFound, ImportError):  # pragma: no cover
+        pass  # pragma: no cover
--- a/geoalchemy2/__init__.py
+++ b/geoalchemy2/__init__.py
@@ -449,11 +449,19 @@ __version__ = "UNKNOWN VERSION"
 # back to pkg_resources for Python 3.7 support
 try:
     import importlib.metadata
-    __version__ = importlib.metadata.version('GeoAlchemy2')
 except ImportError:
     try:
         from pkg_resources import DistributionNotFound
         from pkg_resources import get_distribution
-        __version__ = get_distribution('GeoAlchemy2').version
-    except (DistributionNotFound, ImportError):  # pragma: no cover
-        pass  # pragma: no cover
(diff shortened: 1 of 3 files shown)


adrien-berchet:
Hi @jacob-indigo 
Thank you very much for this work! I didn't know this issue, that's interesting.

For the implementation details, I think we can group everything into one `try` block, something like:
```python
try:
    import importlib.metadata
    __version__ = importlib.metadata.version('GeoAlchemy2')
except ImportError:
    try:
        from pkg_resources import DistributionNotFound
        from pkg_resources import get_distribution
        __version__ = get_distribution('GeoAlchemy2').version
    except:  # pragma: no cover
        pass
```
WDYT?

jacob-indigo:
@adrien-berchet That works for me! Just pushed that change.

adrien-berchet:
After some more thinking, I think we have to write this if we want to handle all cases without a bare Exception:
```python
try:
    import importlib.metadata
except ImportError:
    try:
        from pkg_resources import DistributionNotFound
        from pkg_resources import get_distribution
    except ImportError:  # pragma: no cover
        pass
    else:
        try:
            __version__ = get_distribution('GeoAlchemy2').version
        except DistributionNotFound:  # pragma: no cover
            pass
else:
    try:
        __version__ = importlib.metadata.version('GeoAlchemy2')
    except importlib.metadata.PackageNotFoundError:  # pragma: no cover
        pass
```
Do you agree? Did I miss anything again?

adrien-berchet:
LGTM now.
Is it fine for you @jacob-indigo ?

jacob-indigo:
@adrien-berchet Good catch on the missing `PackageNotFoundError` except. lgtm ✅

adrien-berchet:
Thanks!

PR #400, opened 2022-08-23
Restore compatibility with SQLAlchemy >= 1.1; < 1.4

adrien-berchet:


adrien-berchet:
Some parts of this PR were ported to #425 but the compatibility with `SQLAlchemy >= 1.1, 1.4` is not officially restored.

PR #418, opened 2023-01-03
Remove some irrelevant warnings for Raster type

### LINKED ISSUE #417: Remove warning about srid not being enforced when geometry_type is None
I think there should be no warning about srid not being enforced when geometry_type is None, 
I think these 2 lines should be removed:
https://github.com/geoalchemy/geoalchemy2/blob/3337e5e7580a80bb500814fa1f2ef3a32f317673/geoalchemy2/types.py#L247-L248

If there is no geometry, it is obvious that srid is not enforced. The warnings pollute the output, for example when running pytest (the warnings can be suppressed in pytest, but it would be clearer if they were not output at all).

adrien-berchet:
Fixes #417

### DIFF
--- a/geoalchemy2/types.py
+++ b/geoalchemy2/types.py
@@ -378,6 +378,10 @@ class Raster(_GISType):
         )
         self.extended = None
 
+    @staticmethod
+    def check_ctor_args(*args, **kwargs):
+        return None, -1
+
 
 class _DummyGeometry(Geometry):
     """A dummy type only used with SQLite."""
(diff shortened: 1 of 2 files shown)


adrien-berchet:
Could you test this branch @jwkaltz plz?

jwkaltz:
Hi @adrien-berchet 
The code looks correct; in our project test cases (pyramid_oereb) however we don't have this situation. I will check more closely how we are producing the test data which leads to this warning.

adrien-berchet:
Ok, thanks.
So I guess you create Geometry or Geography columns with no geometry type but with a specific SRID, which might lead to unexpected behavior since the SRID is ignored in this case (or maybe this behavior changed in PostGIS in the last few years?).

PR #421, opened 2023-01-23
Fix: Use empty event for unsupported dialects

### LINKED ISSUE #420: SQLAlchemy event handlers are too aggressive
When [updating](https://github.com/ibis-project/ibis/pull/5296) our `geoalchemy2` dependency to the latest version over in [the ibis project](https://github.com/ibis-project/ibis), we start to see [CI test failures](https://github.com/ibis-project/ibis/actions/runs/3986865996/jobs/6835952644#step:14:1363).

The [important bit](https://github.com/ibis-project/ibis/actions/runs/3986865996/jobs/6835952644#step:14:1427) shows that geoalchemy2 is now registering global event handlers for various DDL operations at import time and those handlers are enforcing a particular dialect.

This means that a user cannot do the following:

1. install `geoalchemy2`
2. use a dialect that doesn't support geometry/geography columns (e.g., `mysql`)
3. run a SQLAlchemy DDL operation using that dialect

It doesn't seem like these handlers are actually doing anything *except* enforcing the dialect, so perhaps they should be removed.

adrien-berchet:
Fixes #420

### DIFF
--- a/geoalchemy2/__init__.py
+++ b/geoalchemy2/__init__.py
@@ -7,6 +7,7 @@ from sqlalchemy.sql import func
 
 from geoalchemy2 import functions  # noqa
 from geoalchemy2 import types  # noqa
+from geoalchemy2.dialects import common
 from geoalchemy2.dialects import postgresql
 from geoalchemy2.dialects import sqlite
 from geoalchemy2.dialects.common import _check_spatial_type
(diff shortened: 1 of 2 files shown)


cpcloud:
This PR fixes #420, thank you!

adrien-berchet:
> This PR fixes #420, thank you!

Perfect, thanks!

PR #434, opened 2023-03-28
setup.py needs to know about geoalchemy2.dialects

EdwardBetts:
If geoalchemy2.dialects is missing from the packages line of setup.py then then building will either show a big warning, or miss the geoalchemy2.dialects package entirely.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setup(
     },
     license="MIT",
     python_requires=">=3.7",
-    packages=find_namespace_packages(include=["geoalchemy2"]),
+    packages=find_namespace_packages(include=["geoalchemy2", "geoalchemy2.dialects"]),
     include_package_data=True,
     zip_safe=False,
     setup_requires=["setuptools_scm"],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -29,7 +29,7 @@
     },
     license="MIT",
     python_requires=">=3.7",
-    packages=find_namespace_packages(include=["geoalchemy2"]),
+    packages=find_namespace_packages(include=["geoalchemy2", "geoalchemy2.dialects"]),
```
adrien-berchet, line 32:
```suggestion
    packages=find_namespace_packages(include=["geoalchemy2*"]),
```
It would be simpler to write this since we also have to add `geoalchemy2.admin`, `geoalchemy2.admin.dialects`, `geoalchemy2.types` and `geoalchemy2.types.dialects`.

adrien-berchet (changes requested):
Hi @EdwardBetts 
Indeed I forgot to update this, thanks for this PR!

adrien-berchet:
Thank you very much @EdwardBetts !

PR #438, opened 2023-04-05
Add an example to insert raster entries in gallery

adrien-berchet:


adrien-berchet:
@mht2953658596 you mentioned an error in the related issue, could you elaborate a bit here please? So I can fix it.
Thanks

mht2953658596:
there is no erroe ,thanks

adrien-berchet:
Ah ok, cool!
Thanks for your check!

Issue #442, opened 2023-04-20
Issue with reflection and empty schema in ORM model


EtienneDG:
Hello there,

I tripped upon an issue using MySQL with the latest commit from master.

My context is the following :
I have 2 Mysql databases : `db_dev` and `db_test` (you guessed their usage from their names).
I never specify a shema name in my models, which is fine with SQLAlchemy that uses other ways to get the schema name if needed.
I run `alembic check` on CI to ensure that I don't have diff between my model and an up-to-date dummy DB.

When reflecting a column for MySQL (the code looks similar for postgre, but I don't know if it causes the same issue) GeoAlchemy2  adds an `AND` clause on `INFORMATION_SCHEMA.COLUMNS:SCHEMA_NAME` IF the schema of the table is not None. 

https://github.com/geoalchemy/geoalchemy2/blob/aac6d078d23dc7c6153a9fc13383e925f0dfe7e1/geoalchemy2/admin/dialects/mysql.py#L34-L40

In my case, this causes the `geometry_type_query` to return two lines : 
- one for the column in the `db_dev`
- one for the column in the `db_test`

And further down the line, causes SQLAlchemy to raise this error : 
```
  File "/home/.../lib/python3.10/site-packages/geoalchemy2/admin/__init__.py", line 99, in _reflect_geometry_column
    select_dialect(inspector.bind.dialect.name).reflect_geometry_column(
  File "/home/.../lib/python3.10/site-packages/geoalchemy2/admin/dialects/mysql.py", line 41, in reflect_geometry_column
    geometry_type, srid, nullable_str = inspector.bind.execute(text(geometry_type_query)).one()
  File "/home/.../lib/python3.10/site-packages/sqlalchemy/engine/result.py", line 1258, in one
    return self._only_one_row(
  File "/home/.../lib/python3.10/site-packages/sqlalchemy/engine/result.py", line 614, in _only_one_row
    raise exc.MultipleResultsFound(
sqlalchemy.exc.MultipleResultsFound: Multiple rows were found when exactly one was required
```

A fix could be to default `schema` to `inspector.default_schema_name` if the schema of the current table is `None`, inside the `reflect_geometry_column` method.

I can propose a PR if it helps.

adrien-berchet:
Very interesting, thank you very much for this report!
Indeed, I guess we should always check the default schema, even for other dialects. If you have some time to propose a PR for this it would be very helpful, so you can test it on your own test case directly.

[...]

PR #443, opened 2023-04-24
fix: handle mysql schemas in column reflection

EtienneDG:
Proposed fix for #442 .

In the end I edited only the MySQL part.

Schema management (per DB) for postgre is a bit different and non fully-qualified tables are not always considered to be in `public` schema. 

@adrien-berchet had some trouble wrapping my head around all the fixtures, tried to simplify a bit what was made in the test suite I edited.

### DIFF
--- a/geoalchemy2/admin/dialects/mysql.py
+++ b/geoalchemy2/admin/dialects/mysql.py
@@ -29,6 +29,7 @@ def reflect_geometry_column(inspector, table, column_info):
         return
 
     column_name = column_info.get("name")
+    schema = table.schema or inspector.default_schema_name
 
     # Check geometry type, SRID and if the column is nullable
     geometry_type_query = """SELECT DATA_TYPE, SRS_ID, IS_NULLABLE
@@ -36,8 +37,8 @@ def reflect_geometry_column(inspector, table, column_info):
         WHERE TABLE_NAME = '{}' and COLUMN_NAME = '{}'""".format(
         table.name, column_name
     )
-    if table.schema is not None:
-        geometry_type_query += """ and table_schema = '{}'""".format(table.schema)
+    if schema is not None:
+        geometry_type_query += """ and table_schema = '{}'""".format(schema)
     geometry_type, srid, nullable_str = inspector.bind.execute(text(geometry_type_query)).one()
     is_nullable = str(nullable_str).lower() == "yes"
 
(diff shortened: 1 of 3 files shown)


geoalchemy2/admin/dialects/mysql.py
```diff
@@ -29,15 +29,16 @@ def reflect_geometry_column(inspector, table, column_info):
         return
 
     column_name = column_info.get("name")
+    schema = table.schema or inspector.default_schema_name
```
adrien-berchet, line 32: What happens if `table.schema is None`? Is it allowed by MySQL? Or should we check for `table.schema is not None`?

adrien-berchet (commented):
Thank you very much for this work @EtienneDG !

geoalchemy2/admin/dialects/mysql.py
EtienneDG: MySQL and Postgre differ on what schemas are.
In postgre : server -> databases -> schemas
In mysql : server -> databases (aka schemas)

You can't* create a table outside a database (ie. outside a schema) in MySQL. If the table has no schema in the ORM model, by default it will be created on the database specified in the connection string (ie. the one in `default_schema_name`).

So if `table.schema` is falsy we use the `default_schema_name`.

It is possible to verify what happens for `''` with this snippet :
```python
t = Table("lake", MetaData(schema=''), autoload_with=conn)
# raises an exception
# (1102, "Incorrect database name ''")
# [SQL: SHOW CREATE TABLE ``.`lake`]
```


*[you can but it involves some specifics](https://dev.mysql.com/doc/refman/8.0/en/innodb-create-table-external.html)

adrien-berchet: Ok, I see. Thanks for the explanations!

tests/test_functional_mysql.py
```diff
@@ -329,26 +329,38 @@ def test_insert(self, conn, Lake, setup_tables):
...
+            with connection.begin():
+                connection.execute(drop)
+                connection.execute(text(f"CREATE DATABASE {temp_db_name};"))
+
+        engine = create_engine(f"mysql://gis:gis@localhost/{temp_db_name}")
+
+        with engine.connect() as connection:
+            with connection.begin():
+                reflection_tables_metadata.drop_all(connection, checkfirst=True)
+                reflection_tables_metadata.create_all(connection)
```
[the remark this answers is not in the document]
EtienneDG: Looks much better indeed

adrien-berchet:
Thank you very much @EtienneDG !

Issue #410, opened 2022-11-30
Adding Geopackage support


caspervdw:
Hi @adrien-berchet , first of all, thanks for maintaining this very useful package. We use it intensively to manage our spatialite files.

Recently I investigated whether we could also interface geopackage files with sqlalchemy + geoalchemy. After some tinkering I found that by calling `SELECT EnableGpkgAmphibiousMode()`, you can actually use geoalchemy as-is to use the ORM to read/write in existing geopackage files. Apparently the spatialite driver has support for geopackage (see https://www.gaia-gis.it/gaia-sins/spatialite-sql-5.0.1.html#p16gpkg)

See here for my implementation: https://github.com/nens/threedi-modelchecker/blob/1eca28ca8854d68df1f4fa136903ce15ea0d9701/threedi_modelchecker/threedi_database.py#L64

As geopackage is quite a common format nowadays, I think it might be interesting for geoalchemy to make this a fully supported option. Stuff that I can see that would need attention:

- discerning geopackage and spatialite dialects and calling the `EnableGpkgAmphibiousMode()`
- inspecting geometry columns and indexes 
- adding / removing geometry columns 
- adding / removing spatial indexes

Please let me know what you think about this. I also might be able to help out here in implementing or testing.

adrien-berchet:
Hi @caspervdw 
That's a great idea! Thanks for introducing me to this feature, I didn't know that.
I think we should solve #395 first so then we can add this new feature in the appropriate module. I will try to work on it soon so then we can work on what you propose.

adrien-berchet:
Hey
I took some time to reorganize the code in #414, so the dialects-specific stuffs are separated. So now we can add support for GeoPackage into `geoalchemy2.dialects.sqlite`. I will not have time to work on this before the end of the year (at least) but if anyone is willing to start a PR on this topic I will be glad to review it and help.

adrien-berchet:
Hi @caspervdw 
Sure, any help is always welcome :)
I will send you an email to speak about this soon.

adrien-berchet:
Solved by #456

PR #456, opened 2023-06-25
Add GeoPackage support

adrien-berchet:
Solves #410 

This PR adds support of GeoPackage using Spatialite.

Main things to check:
- [ ] Should we automatically activate the Amphibious mode?
- [ ] Should we automatically activate the AutoGPKG mode?
- [x] How to use ST_Transform with a GeoPackage? Should we add a specific function to create the required SRS table?


adrien-berchet:
@caspervdw could you try this version please?

tests/test_functional_sqlite.py
```diff
@@ -84,15 +229,90 @@ def test_all_indexes(self, conn, TableWithIndexes, setup_tables):
...
+                    SELECT
+                      srs_id AS srid,
+                      organization AS auth_name,
+                      organization_coordsys_id AS auth_srid,
+                      definition AS srtext,
+                      NULL
+                    FROM gpkg_spatial_ref_sys AS A
+                    WHERE NOT EXISTS (SELECT srid FROM spatial_ref_sys WHERE srs_id = A.srs_id);"""
+                )
+            )
```
adrien-berchet, line 285: I was not able to make the `ST_Transform` function without it. IMO it's something that should be fixed on the Spatialite side but until then we could create a function to make it easier for users.

geoalchemy2/admin/dialects/sqlite.py
```diff
@@ -172,29 +275,112 @@ def before_create(table, bind, **kw):
...
+            if _check_spatial_type(col.type, Geometry, dialect) and check_management(
+                col, dialect_name
+            ):
+                col.type = col._actual_type
+                del col._actual_type
+                dimension = get_col_dim(col)
+                has_z = "Z" in dimension
+                has_m = "M" in dimension
+
+                bind.execute(
```
caspervdw, line 327: There is always an "SQL INJECTION" alarm going off in my head if I see string formatting on raw sql. It's probably fine here, but still, it's easy to do it like SQLAlchemy advertises. Unless I am missing something.

```
t = text("SELECT * FROM users WHERE id=:user_id")
result = connection.execute(t, user_id=12)
```

geoalchemy2/admin/dialects/sqlite.py
```diff
@@ -172,29 +275,112 @@ def before_create(table, bind, **kw):
...
+        elif len(gis_cols) == 1:
+            col = gis_cols[0]
+            srid = col.type.srid
+
+            if col.type.geometry_type is None:
+                col.type.geometry_type = "GEOMETRY"
+
+            # Add the SRID of the table in 'gpkg_spatial_ref_sys' if this table exists
+            if (
+                bind.execute(text("""PRAGMA main.table_info("gpkg_spatial_ref_sys");""")).fetchall()
```
caspervdw, line 293: Why "if it exists"? A geopackage without this table is invalid. GDAL also raises if that is the case.

geoalchemy2/admin/dialects/sqlite.py
```diff
@@ -28,19 +44,73 @@ def load_spatialite(dbapi_conn, connection_record):
...
+                    SELECT table_name, column_name, COUNT(*) AS has_index
+                    FROM gpkg_extensions
+                    WHERE table_name = '{table_name}'
+                        AND column_name = '{column_name}'
+                        AND extension_name = 'gpkg_rtree_index'
+                ) AS B
+                ON A.table_name = B.table_name AND A.column_name = B.column_name
+                WHERE A.table_name = '{table_name}' AND A.column_name = '{column_name}';
+            """.format(
+                    table_name=table_name, column_name=col_name
```
caspervdw, line 86: This query here is similar to the one in OGR. Good to compare them:

https://github.com/OSGeo/gdal/blob/6e9103bd5acb1ff8da305c4e77fa30335ffa3b70/ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp#L1594

One thing I see is that OGR puts `lower()` around the table_name. There is also some recommendation in the gpkg spec:

> Implementers should be aware of the fact that SQLite table names are not case sensitive and that table names in sqlite_master and gpkg_extensions may not have the same case. When searching for table name references, it is recommended to transform table names to lower case with the lower() function. See the Abstract Test Suite for an example.

geoalchemy2/admin/dialects/sqlite.py
```diff
@@ -172,29 +275,112 @@ def before_create(table, bind, **kw):
...
+                col.type = col._actual_type
+                del col._actual_type
+                dimension = get_col_dim(col)
+                has_z = "Z" in dimension
+                has_m = "M" in dimension
+
+                bind.execute(
+                    text(
+                        """INSERT INTO gpkg_contents
+                        VALUES (
```
caspervdw, line 330: I think it is always good to explicitly state the column names when inserting. That way you are forwards compatible with possibly added columns.

geoalchemy2/admin/dialects/sqlite.py
```diff
@@ -209,15 +395,55 @@ def after_create(table, bind, **kw):
...
+            # TODO: This is useless but if we remove it then the disable_spatial_index should be
+            # tested separately
+            disable_spatial_index(bind, table, col)
+
+            # Remove metadata from internal tables
+            # (this is equivalent to DiscardGeometryColumn but for GeoPackage)
+            bind.execute(
+                text(
+                    """DELETE FROM gpkg_extensions
+                    WHERE table_name = '{}'
```
caspervdw, line 422: Good to add `lower()` around both sides of the equality, like here:
https://github.com/OSGeo/gdal/blob/6e9103bd5acb1ff8da305c4e77fa30335ffa3b70/ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp#L6644

tests/__init__.py
```diff
@@ -118,13 +123,15 @@ def copy_and_connect_sqlite_db(input_db, tmp_db, engine_echo):
         engine._spatialite_version = 3
     elif input_db.endswith("spatialite_ge_4.sqlite"):
         engine._spatialite_version = 4
+    elif input_db.endswith(".gpkg"):
```
caspervdw, line 126: should make this case-insensitive

tests/test_functional_sqlite.py
```diff
@@ -6,16 +9,118 @@
...
+            MetaData(),
+            Column("id", Integer, primary_key=True),
+            Column("geom", Geometry()),
+            schema="gis",
+        )
+
+        # Create the table
+        t.create(bind=conn)
+
+        # Drop the table
```
caspervdw, line 122: Is there any check here if the table actually exists and has the correct schema? or is that done implicitly by the `drop()`?

tests/test_functional_sqlite.py
caspervdw: I think such a function would be required for our application; we use `ST_Transform` at some points

caspervdw (approved):
Looks good! I'll now proceed with the testing

geoalchemy2/admin/dialects/sqlite.py
```diff
@@ -28,19 +44,73 @@ def load_spatialite(dbapi_conn, connection_record):
...
+        databases = [
+            i[-1].endswith(".gpkg") for i in dbapi_conn.execute("PRAGMA database_list;").fetchall()
+        ]
+        is_GeoPkg = any(databases)
+    except Exception:
+        is_GeoPkg = False
+    if is_GeoPkg:
+        if not dbapi_conn.execute("SELECT CheckGeoPackageMetaData();").fetchone()[0]:
+            # This only works on the main database
+            dbapi_conn.execute("SELECT gpkgCreateBaseTables();")
```
caspervdw, line 57: As `InitSpatialMetadata()` populates the srs table, I think it is good to call `gpkgInsertEpsgSRID()` for the geopackage at this, so that behaviour is the same.

Users that don't want to have the initialization could just roll their own `load_spatialite` handler

geoalchemy2/admin/dialects/sqlite.py
```diff
@@ -17,6 +18,21 @@
 from geoalchemy2.types import _DummyGeometry
 
 
+def is_gpkg(bind):
```
caspervdw, line 21: There are two ways of "identifying" a geopackage now:

- the one over here (by calling the HasGeopackage() and CheckGeoPackageMetaData())
- the one in the `load_spatialite`

For me it has lead to an issue when creating a new geopackage: I make the connection, load the spatialite library myself, then call `gpkgCreateBaseTables` myself, and continue with creating tables. The cached `bind["_is_gpkg"]` is still False in that case, so I run into errors because geoalchemy sees it as a spatialite `RecoverGeometryColumn`.

Why did you opt for not making geopackage a specific dialect? I think it is perfectly fine to expect from users to initiate a dbconn with something like `gpkg:///path/tofile`. Implementation-wise it would clean up a lot also, I expect.

geoalchemy2/admin/dialects/sqlite.py
```diff
@@ -172,29 +275,112 @@ def before_create(table, bind, **kw):
...
+                has_z = "Z" in dimension
+                has_m = "M" in dimension
+
+                bind.execute(
+                    text(
+                        """INSERT INTO gpkg_contents
+                        VALUES (
+                            '{}',
+                            'features',
+                            NULL,
```
caspervdw, line 333: - QGis and fiona require an identifier; as every table can only have 1 geometry column I suggest putting the table name over here.
- The desciption should be an empty string, not NULL

caspervdw:
@adrien-berchet , I tested it and there seems to be a working geopackage generated. Some issues I could solve at my side `threedi-schema`, others I noted in the review. 

One big suggestion is making this a different dialect; it seems logical looking at the code, and the "auto-discovery" gave some issues for me. For the rest it is looking good!

adrien-berchet:
Hi @caspervdw 
Thanks for your feedback and review!
The specific dialect is an interesting idea, as it would solve the different ways of identifying GPKG in `load_spatialite` and in `is_gpkg`. I suspect it's bit overkill but I will see how we can do this.

tests/test_functional_sqlite.py
adrien-berchet: Ok, I will add it then.

geoalchemy2/admin/dialects/sqlite.py
adrien-berchet: Well, in our case we only use attributes that come from SQLAlchemy objects (e.g. `table.name\, `col.name`, etc.). So it's hard for SQL injections to come up to here. But I guess it's not impossible and any way perfs are not really important  for administration functions so let's use parameters to be sure.

geoalchemy2/admin/dialects/sqlite.py
adrien-berchet: Indeed, good point

geoalchemy2/admin/dialects/sqlite.py
adrien-berchet: Indeed I missed this in the gpkg specs, thanks!

geoalchemy2/admin/dialects/sqlite.py
adrien-berchet: Well, then we are no more forward compatible with renamed columns, which actually already happened in SpatiaLite tables (between v3 and v4 if I remember correctly). So I don't really know what's the best option...

geoalchemy2/admin/dialects/sqlite.py
adrien-berchet: :+1:

tests/__init__.py
adrien-berchet: Yeah, why not :)

tests/test_functional_sqlite.py
adrien-berchet: The drop only checks that the table exists, but not the schema. I will add a check for that, it will be better indeed.

geoalchemy2/admin/dialects/sqlite.py
adrien-berchet: I agree it's super long, that's why I use already initialized DBs in the tests.
Using `InitSpatialMetadata(1)` could be interesting, though it would be a breaking change (or we could just let the use pass a parameter to `load_spatialite`).
Anyway, I agree it should be consistent between SpatiaLite and gpkg, so I will see what's the best option.

geoalchemy2/admin/dialects/sqlite.py
adrien-berchet: Ok, let's make it compatible with QGIS, it makes sense.

geoalchemy2/admin/dialects/sqlite.py
adrien-berchet: In the end I added the possibility to pass an `init_mode` to the function, so it is possible to initialize the table with no SRID and then add the relevant ones manually. Unfortunately, it is not possible to pass an argument to the `sqlalchemy.event.listen` function, so if someone wants to use this argument he has to either create a new function that calls `load_spatialite` with an hard-coded `init_mode` or call `load_spatialite` manually after the connection is created.

geoalchemy2/admin/dialects/sqlite.py
adrien-berchet: I created the new dialect. I had to rework many things but it made the code much better and robust, thanks for the suggestion!

geoalchemy2/admin/dialects/sqlite.py
adrien-berchet: I made these changes and now it seems very close to a GeoPackage created by QGIS. Does it look good to you?

tests/conftest.py
```diff
@@ -34,6 +34,11 @@ def pytest_addoption(parser):
         action="store",
         help="SQLite DB URL used for tests with SpatiaLite4 (`sqlite:///path_to_db_file`).",
     )
+    parser.addoption(
+        "--sqlite_GeoPackage_dburl",
```
elisallenens, line 38: this might be more logical as `"--sqlite_geopackage_dburl"`; the other command line options are also fully in lowercase, including terms like SQLite and PostgreSQL

geoalchemy2/types/__init__.py
```diff
@@ -436,6 +437,17 @@ class GeometryDump(CompositeType):
 _sqlite_ischema_names["GEOMETRYCOLLECTION"] = Geometry
 _sqlite_ischema_names["RASTER"] = Raster
 
+# _gpkg_ischema_names["GEOMETRY"] = Geometry
```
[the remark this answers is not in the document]
adrien-berchet: Because I forgot to remove it :innocent: 
As far as I could see it's not required since we inherit from SQLite dialect.

tests/conftest.py
adrien-berchet: Indeed, thanks!

geoalchemy2/admin/dialects/sqlite.py
caspervdw: Yes it certainly looks a lot better now! Great!

caspervdw:
@adrien-berchet  I am :+1: on merging this. Two observations during testing: 

**Alembic compatibility**

To make this work with alembic, you should register it as a dialect:

```
from alembic.ddl.impl import DefaultImpl

class GpkgImpl(DefaultImpl):
    __dialect__ = 'geopackage'
```

**ogr-info complains**

If I run ogr-info on a gpkg generated using this PR I get:

```
Warning 1: Field format 'VARCHAR(50)' not supported
```

Looking at the gpkg spec (section 1.1.1.1.3) there is indeed no mention of a VARCHAR. It should be `TEXT` or `TEXT(50)`. 

These two are no blockers for us however, so, great work!

adrien-berchet:
> **Alembic compatibility**
> 
> To make this work with alembic, you should register it as a dialect:
> 
> ```
> from alembic.ddl.impl import DefaultImpl
> 
> class GpkgImpl(DefaultImpl):
>     __dialect__ = 'geopackage'
> ```

In `geoalchemy2.alembic_helpers` I register it like this:
```python
class GeoPackageImpl(SQLiteImpl):
    """Class to copy the Alembic implementation from SQLite to GeoPackage."""

    __dialect__ = "geopackage"
```
Isn't it enough? Did you import `geoalchemy2.alembic_helpers` in your alembic script? Because this module is not automatically imported with GeoAlchemy2, it has to be explicitly imported.

> **ogr-info complains**
> 
> If I run ogr-info on a gpkg generated using this PR I get:
> 
> ```
> Warning 1: Field format 'VARCHAR(50)' not supported
> ```
> 
> Looking at the gpkg spec (section 1.1.1.1.3) there is indeed no mention of a VARCHAR. It should be `TEXT` or `TEXT(50)`.

I'm not able to reproduce this, do you know to which table and field this warning is related?
Is it only when you use the `geoalchemy2.admin.dialects.geopackage.populate_spatial_ref_sys()` function?


> These two are no blockers for us however, so, great work!

Great, thanks!
I will try to fix these two issues so we can release a clean version soon.

caspervdw:
> @caspervdw I performed a few tests with Alembic and everything seems to work in my case so I don't know what's going on on your side. I still think it's because you did not import `geoalchemy2.alembic_helpers`.

Sorry I indeed didn’t import the helpers. So that’s ok then!

For ogr-info; I used the command line tool on Ubuntu 22.04. It is GDAL 3.4.1 .

adrien-berchet:
> Sorry I indeed didn’t import the helpers. So that’s ok then!

Ok, no problem :)

> For ogr-info; I used the command line tool on Ubuntu 22.04. It is GDAL 3.4.1 .

I'm still not able to reproduce this issue locally but there is only one place where I create `VARCHAR` columns: in `geoalchemy2.admin.dialects.geopackage.populate_spatial_ref_sys()`. So after reading again the GeoPackage specs I created a `spatial_ref_sys` view instead of a table, as described here: http://www.geopackage.org/spec/#_gpkg_spatial_ref_sys
And I renamed the `populate_spatial_ref_sys()` function into `create_spatial_ref_sys_view` to be more consistent. Everything still work for me and there is no `VARCHAR` anywhere so I think we are good now.

caspervdw:
> I'm still not able to reproduce this issue locally but there is only one place where I create `VARCHAR` columns: in `geoalchemy2.admin.dialects.geopackage.populate_spatial_ref_sys()`. So after reading again the GeoPackage specs I created a `spatial_ref_sys` view instead of a table, as described here: http://www.geopackage.org/spec/#_gpkg_spatial_ref_sys And I renamed the `populate_spatial_ref_sys()` function into `create_spatial_ref_sys_view` to be more consistent. Everything still work for me and there is no `VARCHAR` anywhere so I think we are good now.

Very sorry, I tried to reproduce, and it appeared I had a `String(100)` in my table definition. If I replace it with `Text(100)`, the column is neatly created with type `TEXT(100)`. So we're good here.
It might be good though to limit the number of allowed types when using the `gpkg` driver to those specified in the spec (section 1.1.1.1.3). But certainly not required for our purposes.

> In the end I added the possibility to pass an init_mode to the function, so it is possible to initialize the table with no SRID and then add the relevant ones manually. Unfortunately, it is not possible to pass an argument to the sqlalchemy.event.listen function, so if someone wants to use this argument he has to either create a new function that calls load_spatialite with an hard-coded init_mode or call load_spatialite manually after the connection is created.

That's all right, you can do it with a `lambda` like this:

```
listen(engine, "connect", lambda x, y: load_spatialite(x, y, init_mode="EMPTY")
```

adrien-berchet:
> Very sorry, I tried to reproduce, and it appeared I had a `String(100)` in my table definition. If I replace it with `Text(100)`, the column is neatly created with type `TEXT(100)`. So we're good here. It might be good though to limit the number of allowed types when using the `gpkg` driver to those specified in the spec (section 1.1.1.1.3). But certainly not required for our purposes.

Ahah ok, no problem.
For now I'm going to let the users define whatever type they want because GeoAlchemy 2 should only handle spatial types. Such limitation you propose should go to SQLAlchemy IMO.

> That's all right, you can do it with a `lambda` like this:
> 
> ```
> listen(engine, "connect", lambda x, y: load_spatialite(x, y, init_mode="EMPTY")
> ```

Exactly. I improve the docs for this in #457 to be more complete.

Issue #458, opened 2023-07-04
Performance improvement suggestions for load_spatialite


caspervdw:
When creating a new spatialite file, you should call `SELECT InitSpatialMetaData();`. This is done by `load_spatialite`. However this call takes very long (minutes!).

There are several ways for optimizing this:

1. Limit the size of the spatial ref table: (done in #456)
2. Use the optional argument `transaction` and set it to TRUE. Then the whole operation will be handled as a single Transaction (faster): the default setting is transaction=FALSE (slower, but safer). (copy-paste from http://www.gaia-gis.it/gaia-sins/spatialite-sql-latest.html)
3. Set `PRAGMA journal_mode = MEMORY`. This is less safe, but again a lot faster.

We use both 'unsafe' methods and additionally wrap the whole initialization thing in a "file transaction": https://github.com/nens/threedi-schema/blob/master/threedi_schema/application/threedi_database.py#L136 to make it safe.

I think it is good to remove the initalization stuff from the `load_spatialite` and instead, make explicit methods `spatialite_is_initialized() -> bool`, `init_spatialite` which can be called by the user.

adrien-berchet:
Solved by #459

PR #459, opened 2023-07-04
Split load_spatialite into load_spatialite_driver and init_spatialite

adrien-berchet:
Solves #458

### DIFF
--- a/geoalchemy2/admin/dialects/geopackage.py
+++ b/geoalchemy2/admin/dialects/geopackage.py
@@ -38,7 +38,7 @@ registry.register("gpkg", "geoalchemy2.admin.dialects.geopackage", "GeoPackageDi
 
 
 def load_geopackage_driver(dbapi_conn, *args):
-    """Load SpatiaLite extension in GeoPackage connection.
+    """Load SpatiaLite extension in GeoPackage connection and set VirtualGpkg and Amphibious modes.
 
     .. Warning::
         The path to the SpatiaLite module should be set in the `SPATIALITE_LIBRARY_PATH`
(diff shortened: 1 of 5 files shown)


adrien-berchet:
@caspervdw would something like this make sense to you? If yes I will add the transaction and journal arguments to the `init_spatialite` function.

geoalchemy2/admin/dialects/sqlite.py
```diff
@@ -16,25 +17,35 @@
...
+        dbapi_conn: The DBAPI connection.
     """
     if "SPATIALITE_LIBRARY_PATH" not in os.environ:
         raise RuntimeError("The SPATIALITE_LIBRARY_PATH environment variable is not set.")
     dbapi_conn.enable_load_extension(True)
     dbapi_conn.load_extension(os.environ["SPATIALITE_LIBRARY_PATH"])
     dbapi_conn.enable_load_extension(False)
 
+
+def init_spatialite(dbapi_conn, *args, init_mode: Optional[str] = None):
```
caspervdw, line 37: Why adding the `*args` here?

geoalchemy2/admin/dialects/geopackage.py
```diff
@@ -64,6 +71,17 @@ def load_spatialite_gpkg(dbapi_conn, connection_record):
     dbapi_conn.execute("SELECT EnableGpkgAmphibiousMode();")
```
caspervdw, line 71: I think this line (`EnableGpkgAmphibiousMode`) should go to `load_geopackage_driver` as it should be called *always* rather than just on initialization of the file.

caspervdw (commented):
Good improvement, 2 suggestions!

geoalchemy2/admin/dialects/sqlite.py
adrien-berchet: Because when it is called as a listener it receives another argument that we don't use.

geoalchemy2/admin/dialects/geopackage.py
adrien-berchet: Good point, I moved it.

adrien-berchet:
@caspervdw I added the 2 other parameters (`transaction` and `journal_mode`) and reworked the docs accordingly so I think it's good now.

PR #464, opened 2023-07-07
Add type annotations

shuttle1987:
These changes add in type annotations to the code.


geoalchemy2/elements.py
```diff
@@ -114,10 +117,10 @@ class WKTElement(_SpatialElement):
 
     _REMOVE_SRID = re.compile("(SRID=([0-9]+); ?)?(.*)")
 
-    geom_from = "ST_GeomFromText"
-    geom_from_extended_version = "ST_GeomFromEWKT"
+    geom_from: str = "ST_GeomFromText"
+    geom_from_extended_version: str = "ST_GeomFromEWKT"
 
-    def __init__(self, data, srid=-1, extended=None):
+    def __init__(self, data, srid: int = -1, extended: Optional[bool] = None) -> None:
```
adrien-berchet, line 123: Why not adding type annotation for `data`? While under certain conditions (i.e. when `extended=False` of when `extended=True and srid != -1`) it can actually work with anything that can implicitly cast to `str`, it is strongly advised to pass an actual `str` object as `data` argument.

geoalchemy2/elements.py
```diff
@@ -1,6 +1,9 @@
+from __future__ import annotations
```
adrien-berchet, line 1: Why is it needed? I don't see any use of this anywhere. Is it to support Py37?

adrien-berchet (commented):
Very nice, thanks for this PR!!!

adrien-berchet:
Another thing: it could be nice to add some tests with mypy to check the annotations are correct.

geoalchemy2/elements.py
shuttle1987: There's situations where the return type of methods are the same as the class type being defined. For example:

```python
class WKTElement(_SpatialElement):
    def as_wkt(self) -> WKTElement:
        if self.extended:
            srid_match = self._REMOVE_SRID.match(self.data)
            return WKTElement(srid_match.group(3), self.srid, extended=False)
        return WKTElement(self.data, self.srid, self.extended
```
This would cause a circular reference unless we use the newer feature that defers the evaluation of the type annotations until after the classes are defined. See https://peps.python.org/pep-0563/ for more details.

In Python 3.11 we could use the new `Self` annotation type: https://peps.python.org/pep-0673/, but the reason I didn't do it this way is because there's older versions of Python to support.

geoalchemy2/elements.py
shuttle1987: I wasn't sure what exactly to do here with the type of `data` since the following code seems to suggest that it must be a `str` type:
```python
class WKTElement(_SpatialElement):
    """Instances of this class wrap a WKT or EWKT value.

    Usage examples::

        wkt_element_1 = WKTElement('POINT(5 45)')
        wkt_element_2 = WKTElement('POINT(5 45)', srid=4326)
        wkt_element_3 = WKTElement('SRID=4326;POINT(5 45)', extended=True)
    """

    _REMOVE_SRID = re.compile("(SRID=([0-9]+); ?)?(.*)")

    geom_from: str = "ST_GeomFromText"
    geom_from_extended_version: str = "ST_GeomFromEWKT"

    def __init__(self, data, srid: int = -1, extended: Optional[bool] = None) -> None:
        if extended is None:
            extended = data.startswith("SRID=")
        if extended and srid == -1:
            # read srid from EWKT
            data_s = data.split(";")
            if len(data_s) != 2:
                raise ArgumentError("invalid EWKT string {}".format(data))
            header = data_s[0]
            try:
                srid = int(header[5:])
            except ValueError:
                raise ArgumentError("invalid EWKT string {}".format(data))
        _SpatialElement.__init__(self, data, srid, extended)
```

In this case whatever the type is for the `data` parameter it must support the `.startswith` and `.split` methods which seems to indicate that it needs to just be a `str`. We can add a direct annotation on `str` if that works.

adrien-berchet: Yeah for `WKTElement` we should only use `str` objects.
For `WKBElement` I think we can accept `str` and `memoryview`.

geoalchemy2/elements.py
adrien-berchet: Ah ok ok, thanks for the explanation. Let's keep it as is then.

geoalchemy2/elements.py
shuttle1987: See commits 3be4ed5 and b40ca22

adrien-berchet: Perfect, thanks!

adrien-berchet:
Also, I think we have to add a `py.typed` file, right? https://peps.python.org/pep-0561/#packaging-type-information

shuttle1987:
As far as I understand `py.typed` is for when you are distributing type stub files. Because we aren't supporting Python 2 I don't know if we need this?

shuttle1987:
I'm not so familiar with the CI pipeline being used for this project but I'm curious as to why the linting step in tox fails only for python==3.8?

adrien-berchet:
> As far as I understand `py.typed` is for when you are distributing type stub files. Because we aren't supporting Python 2 I don't know if we need this?

I'm not sure either, I never setup type hints for a library that will be used in other codes. Did you try to run Mypy on another code that includes GeoAlchemy2 objects?

> I'm not so familiar with the CI pipeline being used for this project but I'm curious as to why the linting step in tox fails only for python==3.8?

It's just because the lint tox job is run only in the Py38 CI job.

shuttle1987:
That makes sense about the single job failing if that's the only place the liniting is run. I've pushed a couple of commits to attempt to deal with the linter complaints

adrien-berchet:
Sorry I should have tell you that before but you can run `tox r -e format` locally to fix most of the simple linting issues.

adrien-berchet:
Now it's good, thanks :⁠-⁠)
Do you want to add a few tests for these annotations? Or I can do it later if you prefer. I never tried it but I think we can use this pytest plugin: https://pypi.org/project/pytest-mypy-testing/ . But it can be something else if you know a better solution.

shuttle1987:
I'm not familiar with that package, in other projects I've added a mypy check as part of the CI run.

One other question I had was the types found in geoalchemy2/comparator.py, do we want to annotate return types there? I wasn't sure exactly what those types were supposed to be so I didn't annotate them in this PR.

geoalchemy2/elements.py
```diff
@@ -297,9 +307,9 @@ class RasterElement(_SpatialElement):
     most cases you won't need to create ``RasterElement`` instances yourself.
     """
 
-    geom_from_extended_version = "raster"
+    geom_from_extended_version: str = "raster"
 
-    def __init__(self, data):
+    def __init__(self, data) -> None:
```
shuttle1987, line 312: What type should we use to annotate `data` here?

adrien-berchet: I think it should be `str` and `memoryview`

adrien-berchet:
For comparators I'm not sure. We should check the type annotations in SQLAlchemy and do the same.

geoalchemy2/elements.py
shuttle1987: See 2f7b1ad48df96c51ef7e59dc834918b0e29ad668

adrien-berchet:
> For comparators I'm not sure. We should check the type annotations in SQLAlchemy and do the same.

It seems that the comparators can get a WKT `str`, a `WKTElement`, or a `WKBElement` and they return a `sqlalchemy.sql.expression.BinaryExpression`.

shuttle1987:
I'm not so sure how best to annotate the types of the Comparator class, are these changes worth merging as is? Or is anyone able to push a commit with a fix?

adrien-berchet:
As far as I can see, the comparators should return `sqlalchemy.sql.elements.ColumnElement`. And their inputs should be `Union[str, WKTElement]` (and maybe also `WKBElement` but I'm not sure?)
Also, could you try to add tests using `pytest-mypy` please? So we are (almost) sure that annotations are correct.

shuttle1987:
There was some missing type stub files in the last run, I've attempted to install those in https://github.com/geoalchemy/geoalchemy2/pull/464/commits/98d592e399f953012b5abbc26342f28c0813e585

I think there's a few things that will fail however, but I'd like to see what those are from the CI runner. I suspect that we will need to add to the tox configuration some mypy options including the following:

```
# MYPY
[tool.mypy]
ignore_missing_imports = true
```

adrien-berchet:
> There was some missing type stub files in the last run, I've attempted to install those in [98d592e](https://github.com/geoalchemy/geoalchemy2/commit/98d592e399f953012b5abbc26342f28c0813e585)

Nice!

> I think there's a few things that will fail however, but I'd like to see what those are from the CI runner. I suspect that we will need to add to the tox configuration some mypy options including the following:
> 
> ```
> # MYPY
> [tool.mypy]
> ignore_missing_imports = true
> ```

Yeah indeed we will probably have to add some custom config for mypy.

shuttle1987:
Looks like my idea of using `mypy --install-types --non-interactive` isn't going to work nicely inside a CI run since there's no cache.

shuttle1987:
Looks like we encounter this issue: https://github.com/sqlalchemy/sqlalchemy2-stubs/issues/242

adrien-berchet:
> Looks like we encounter this issue: [sqlalchemy/sqlalchemy2-stubs#242](https://github.com/sqlalchemy/sqlalchemy2-stubs/issues/242)

Since this issue is supposed to be fixed in `SQLAlchemy>=2` I suggest we run Mypy only for the `*-sqlalatest` jobs. WDYT?

shuttle1987:
If you could change the mypy test run to only run vs SQLAlchemy >2 that would be great.

adrien-berchet:
In the end I could make it work for both `sqlalchemy==1.4.*` and `sqlalchemy>=2`.
I fixed several things to make Mypy happy and I added some tests for comparators to be sure. Does it look good to you?

geoalchemy2/comparator.py
```diff
@@ -92,15 +94,15 @@ def __getattr__(self, name):
         func_ = _FunctionGenerator(expr=self.expr)
         return getattr(func_, name)
 
-    def intersects(self, other):
+    def intersects(self, other: _COMPARATOR_INPUT_TYPE) -> ColumnElement:
         """The ``&&`` operator. A's BBOX intersects B's."""
         return self.operate(INTERSECTS, other, result_type=sqltypes.Boolean)
 
-    def overlaps_or_to_left(self, other):
+    def overlaps_or_to_left(self, other: _COMPARATOR_INPUT_TYPE):
```
shuttle1987, line 101: Can this return type be annotated as `ColumnElement`?

geoalchemy2/comparator.py
```diff
@@ -92,15 +94,15 @@ def __getattr__(self, name):
...
         """The ``&<`` operator. A's BBOX overlaps or is to the left of B's."""
         return self.operate(OVERLAPS_OR_TO_LEFT, other, result_type=sqltypes.Boolean)
 
-    def overlaps_or_to_right(self, other):
+    def overlaps_or_to_right(self, other: _COMPARATOR_INPUT_TYPE):
```
shuttle1987, line 105: Can this return type be annotated as `ColumnElement`?

shuttle1987 (commented):
Overall this looks good to me.

geoalchemy2/comparator.py
adrien-berchet: Oh yes you're right, I forgot this one. Thanks

adrien-berchet:
Thank you very much for this work @shuttle1987 !
I think I will push a new release soon to make this available.

PR #477, opened 2023-10-08
Fix: Add missing py.typed file to enable static typing

### LINKED ISSUE #476: mypy: Skipping analyzing "geoalchemy2.functions": module is installed, but missing library stubs or py.typed marker  [import]
After updating to GeoAlchemy2 0.14.1 which mentions type annotations in the release notes, I tried removing the ignore_missing_imports exception in my pyproject.toml. Here's the output from running mypy on my source:
```
file.py:4: error: Skipping analyzing "geoalchemy2.functions": module is installed, but missing library stubs or py.typed marker  [import]
file.py:4: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
```
Are the type annotations not meant for use by third parties, so this is as expected?

If it's not as expected, I can say I haven't done anything special for pytest, and mypy somehow figures out the types for that without giving me any warnings. A difference I can see is that there is a py.typed file in that tree: https://github.com/pytest-dev/pytest/tree/main/src/pytest (which was alluded to in https://github.com/geoalchemy/geoalchemy2/pull/464#issuecomment-1625967236).

adrien-berchet:
Fixes #476

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,7 @@ setup(
         "Intended Audience :: Information Technology",
         "License :: OSI Approved :: MIT License",
         "Topic :: Scientific/Engineering :: GIS",
+        "Typing :: Typed",
     ],
     keywords="geo,gis,sqlalchemy,orm",
     author="Eric Lemoine",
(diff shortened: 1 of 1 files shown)


adrien-berchet:
Superseded by #478

PR #499, opened 2024-03-13
Update postgresql.py

KAMIKADZ4:
Feat: Converting types to str

### Description
Postgis allows user to create geometry using a string containing geojson. In SQLAlchemy there is no option to add a type converter for asyncpg, so when writing to the database geojson, which is represented by dict type, an error appears. I decided that converting any types to str is the right solution

### DIFF
--- a/geoalchemy2/types/dialects/postgresql.py
+++ b/geoalchemy2/types/dialects/postgresql.py
@@ -25,4 +25,7 @@ def bind_processor_process(spatial_type, bindvalue):
     elif isinstance(bindvalue, RasterElement):
         return "%s" % (bindvalue.data)
     else:
-        return bindvalue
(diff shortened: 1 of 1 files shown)


adrien-berchet:
Hi @KAMIKADZ4 
Thanks for your contribution!
Could you just give more details about which issue this PR is supposed to fix please? And maybe an example if possible?
As far as I know, you can insert GeoJSON entries with the function `ST_GeomFromGeoJSON`. Isn't it possible to use it in your case? Maybe you just have to convert you dict entries into JSON strings?

adrien-berchet:
I close this PR for now because I don't understand which issue it is supposed to solve and I had no answer from the author.
Feel free to reopen to discuss it.

PR #500, opened 2024-03-28
Allow geometry_type to be None at the type level to match documentation

logankaser:
### Description
Allow geometry_type to be None at the type level to match documentation.
Fixes a common type checking error when using geoalchemy

### Checklist

This pull request is:

- [x] A short code fix

### DIFF
--- a/geoalchemy2/types/__init__.py
+++ b/geoalchemy2/types/__init__.py
@@ -114,7 +114,7 @@ class _GISType(UserDefinedType):
 
     def __init__(
         self,
-        geometry_type="GEOMETRY",
+        geometry_type: Optional[str] = "GEOMETRY",
         srid=-1,
         dimension=2,
         spatial_index=True,
--- a/geoalchemy2/types/__init__.py
+++ b/geoalchemy2/types/__init__.py
@@ -7,7 +7,6 @@ columns/properties in models.
 import warnings
 from typing import Any
 from typing import Dict
-from typing import Literal
 from typing import Optional
 
 from sqlalchemy.dialects import postgresql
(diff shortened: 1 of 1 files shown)


adrien-berchet:
Hi @logankaser 
The other parameters of this function don't cause any issue? Because there are other parameters that are optional so I'm wondering.

logankaser:
I think we just were not using the other arguments on my team, hopefully these types are accurate as
I've not used those arguments.

geoalchemy2/types/__init__.py
```diff
@@ -114,14 +115,14 @@ class _GISType(UserDefinedType):
...
-        geometry_type="GEOMETRY",
+        geometry_type: Optional[str] = "GEOMETRY",
         srid=-1,
         dimension=2,
         spatial_index=True,
         use_N_D_index=False,
-        use_typmod=None,
-        from_text=None,
-        name=None,
+        use_typmod: Optional[Any] | Literal[False] = None,
```
adrien-berchet, line 123: Why not just `Optional[bool]` here?

adrien-berchet:
> I think we just were not using the other arguments on my team, hopefully these types are accurate as I've not used those arguments.

Ok, I see :)
Thanks for the new change!

geoalchemy2/types/__init__.py
logankaser: Technically, it's more accurate, but I did not realize it is not available in `3.7`, bool it is!

adrien-berchet:
Thank you very much!
I'm going to push a new release to make these change available.

adrien-berchet:
The new release is available: https://pypi.org/project/GeoAlchemy2/0.14.7

PR #509, opened 2024-04-29
Fix: Default SRID is bypassed when using floating point coordinates.

aballet:
### Description
This is a very simple one-character fix. Since v0.15.0, the new `SPLIT_WKT_PATTERN` regex did not account for floating point coordinates. The match() method returns None, and the `default_srid` is never applied in `format_geom_type()`.

How do you suggest testing for this? Current functional tests do not have any floating point, so I could add one there? Or make them all floats instead of integers?

### Checklist

This pull request is:

- [x] A short code fix
	- [x] Please include tests that fail with the `main` branch and pass with the provided fix.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,12 @@
 GeoAlchemy 2 Changelog
 ======================
 
+0.15.0
+------
+
+* Specific process for geometries with Z or M coordinate with SpatiaLite dialect @spd5 (#506)
+* Chore: Fix type hints on stubs @adrien-berchet (#504)
+
 0.14.7
 ------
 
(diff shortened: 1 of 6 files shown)


adrien-berchet:
Hi @aballet 
You're absolutely right, thank you very much for spotting this issue and sorry about this stupid miss...
I think it makes sense to update the tests to include floating point numbers, especially in `test_functional.py::TestInsertionCore::test_insert_all_geom_types`. You can add a `pytest.mark.parametrize` to test with only integers or with floating points.

adrien-berchet:
I updated the test in https://github.com/aballet/geoalchemy2/pull/1/ so you can merge and then I'll merge this PR.

aballet:
Ah, I see that I forgot to escape the dot myself, thank you!
I merged your code to include the tests.

adrien-berchet:
Thanks for this work @aballet !
I'm going to push a new release.

adrien-berchet:
The new release is available: https://pypi.org/project/GeoAlchemy2/0.15.1/
I hope it will solve your issue @aballet 
Anyway, thanks again for this PR @aballet !

PR #542, opened 2025-04-20
fix: handle geometryzm case in sqlite like postgis

imnotjames:
### Description

in the postgis dialect, the coord dimensions
for GEOMETRYZM or similar types is overwritten
by the value expected

this does the same thing for spatialite to keep
a similar behavior

### Checklist

This pull request is:

- [ ] A short code fix

### DIFF
--- a/geoalchemy2/admin/dialects/sqlite.py
+++ b/geoalchemy2/admin/dialects/sqlite.py
@@ -169,16 +169,24 @@ def _setup_dummy_type(table, gis_cols):
 
 def get_col_dim(col):
     """Get dimension of the column type."""
-    if col.type.dimension == 4:
-        dimension = "XYZM"
-    elif col.type.dimension == 2:
-        dimension = "XY"
-    else:
-        if col.type.geometry_type.endswith("M"):
-            dimension = "XYM"
-        else:
-            dimension = "XYZ"
-    return dimension
+    coord_dimension = col.type.dimension
+    geometry_type = col.type.geometry_type
+
+    if geometry_type.endswith("ZM"):
+        coord_dimension = 4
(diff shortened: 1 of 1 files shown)


adrien-berchet:
Thank you very much for your proposal @imnotjames !

This way could indeed work in your specific case but the goal of `GeoAlchemy2` is to use the same syntax for all dialects, so the definition of a column type should be processed the same way for all dialects. Nevertheless, I agree that it makes sense to improve it.
I think the best way is to improve the [_GISType.check_ctor_args](https://github.com/geoalchemy/geoalchemy2/blob/master/geoalchemy2/types/__init__.py#L182) method, which is called in the `_GISType` initializer. In this method we could ensure the `geometry_type` and `dimension` values are consistent. For example, if the geometry type is a `M`, `Z` or `ZM` flavor, we could update the `dimension` value accordingly. And the default value for `dimension` could be set to `None`, so if it's not given we compute the value from the given value of `geometry_type` and if it's given we check the consistency of the `geometry_type` and `dimension` values (e.g. `POINTZ` with `dimension==4` is not consistent and should raise a `ValueError` exception).
After that, the value will be consistent and will be properly used in `get_col_dim` that you updated.

Does this plan look good to you?

adrien-berchet:
Hi @imnotjames 
I pushed #544 to show what I meant. As far as I can see it works for your snippet. Can you test on your side to be sure it will solve your issue please?

imnotjames:
> Hi @imnotjames I pushed #544 to show what I meant. As far as I can see it works for your snippet. Can you test on your side to be sure it will solve your issue please?

I'll close this one and take a look later today!  It probably will resolve it for me :)

PR #566, opened 2025-08-01
Feat: Add cockroachdb support

davidshepherd7:
### Description

Cockroachdb is a distributed database which is [mostly postgis-compatible](https://www.cockroachlabs.com/docs/stable/query-spatial-data).

The [sqlalchemy driver for cockroachdb](https://github.com/cockroachdb/sqlalchemy-cockroachdb) provides a dialect based on (inheriting from) the postgres dialect, but with a different name. 

So we can get good cockroachdb support just by either including cockroachdb in the list of dialect names or by using isinstance instead of name string comparison.

### Checklist

This pull request is:

- [ ] A short code fix
- [x] A new feature implementation or update an existing feature

### DIFF
--- a/geoalchemy2/types/__init__.py
+++ b/geoalchemy2/types/__init__.py
@@ -42,6 +42,7 @@ def select_dialect(dialect_name):
         "mysql": dialects.mysql,
         "mariadb": dialects.mariadb,
         "postgresql": dialects.postgresql,
+        "cockroachdb": dialects.postgresql,  # Cockroachdb is postgis-compatible
         "sqlite": dialects.sqlite,
     }
     return known_dialects.get(dialect_name, dialects.common)


davidshepherd7:
Do you think this should have tests? 

Since cockroachdb is postgis compatible it should all "just work" to the extent of that compatibility, and setting up + maintaining this test is relatively hard since it's a whole new database. So I think it's probably not worth your time to maintain it.

FWIW I'm trying out this patch in a large codebase that we're migrating from postgres to cockroachdb and the relevant unittests are passing. But we don't use a lot of GIS functionality.

davidshepherd7:
Hi, thanks! 

> I'm afraid there may be a few issues, especially for reflection for example.

Yeah I think you might be right, on closer inspection some of our remaining failing tests look like they are geoalchemy-related.

I'll have a go at setting up tests. My plan is roughly:

Build a cockroachdb test container, following the mariadb example (I guess you don't want to make it harder to build the main postgres container if something goes wrong with cockroachdb). Then run `test_functional_postgresql.py` against the new container and see what happens.

I'm not sure when I'll work on this, might not be immediately but should be some time this year.

adrien-berchet:
Hi @davidshepherd7 !
Ok, too bad :)
Could you just explain me what you tried and what was the issues in your attempts? I will try to find some time to try on my side (but I don't know when, I'll ping you when it will be possible to test).
Anyway, thank you for your work on this!

